PR #5, opened 2023-02-09
Address linting and deprecated ioutil pkg.

HumairAK:
This addresses the failure [here](https://github.com/red-hat-data-services/data-science-pipelines-operator/pull/4#issuecomment-1424305639)

Donnu why it wasn't caught in upstream repo, but the error was accurate. We might need to make sure the yaml lint config is being picked up in the github action.

There's also a commit that addresses deprecated package for ioutil. Again not sure why this wasn't picked up upstream. My suspicion is both the yamllint and golanglint configs are not being included in the github action.

### DIFF
--- a/.github/pull.yml
+++ b/.github/pull.yml
@@ -1,0 +1,8 @@
+version: "1"
+rules:
+  - base: main
+    upstream: opendatahub-io:main
+    mergeMethod: none  # don't automatically merge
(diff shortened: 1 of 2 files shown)


HumairAK:
one sec, the pull is not supposed to be on here, that change should only go downstream

PR #6, opened 2023-02-09
Use io from io pkg instead of deprecated ioutil.

HumairAK:
This fixes the staticheck pre-commit test failure that occurs locally, not sure why it's ignored via the github action, but will follow up a fix on that next.


anishasthana (approved):
/lgtm

PR #8, opened 2023-02-10
Propose new cr spec.

HumairAK:
Proposal for updated CR Spec. Please provide your thoughts on what else is required here, and we will follow up with code chances to accomodate the new sample spec. 

The `deploy: true` is to allow us to disable components for development purposes, for example, we may want to deploy apiserver locally while wanting to deploy the rest of the stack via operator.

I'm thinking the most minimal spec would be: 

```yaml
apiVersion: dspipelines.opendatahub.io/v1alpha1
kind: DSPipeline
metadata:
  name: sample
spec:
  apiServer:
    deploy: true
  persistentAgent:
    deploy: true
  scheduledWorkflow:
    deploy: true
  viewerCRD:
    deploy: true
  mlpipelineUI:
    deploy: true
  database:
    mariaDB:
      deploy: true
  storage:
    minio:
      deploy: true
```

And everything else is optional with default values.

### DIFF
--- a/config/samples/dspipelines.io_v1alpha1_dspipeline.yaml
+++ b/config/samples/dspipelines.io_v1alpha1_dspipeline.yaml
@@ -2,49 +2,110 @@ apiVersion: dspipelines.opendatahub.io/v1alpha1
 kind: DSPipeline
 metadata:
   name: sample
+  namespace: data-science-project
 spec:
   apiServer:
+    deploy: true
+    image: quay.io/modh/odh-ml-pipelines-api-server-container:v1.18.0-8
+    resources:
+      requests:
+        cpu: 250m
+        memory: 500Mi
+      limits:
+        cpu: 500m
+        memory: 1Gi
     # optional (default is: ds-pipeline-artifact-script-${metadata.name} / artifact_script)
     artifactScriptConfigMap:
       name: ds-pipeline-artifact-script-sample
       key: "artifact_script"
-    apiServerImage: quay.io/modh/odh-ml-pipelines-api-server-container:v1.18.0-8
     artifactImage: quay.io/modh/odh-ml-pipelines-artifact-manager-container:v1.18.0-8
+    dBConfigConMaxLifetimeSec: 120
+    collectMetrics: true
   persistentAgent:
+    deploy: true
     image: quay.io/modh/odh-ml-pipelines-persistenceagent-container:v1.18.0-8
     pipelineAPIServerName: apiserver
+    numWorkers: 2  # Number of worker for sync job.
+    resources:
+      requests:
+        cpu: 120m
+        memory: 500Mi
+      limits:
+        cpu: 250m
+        memory: 1Gi
   scheduledWorkflow:
+    deploy: true
     image: quay.io/modh/odh-ml-pipelines-scheduledworkflow-container:v1.18.0-8
+    resources:
+      requests:
+        cpu: 120m
+        memory: 100Mi
+      limits:
+        cpu: 250m
+        memory: 250Mi
   viewerCRD:
+    deploy: true
     image: quay.io/modh/odh-ml-pipelines-viewercontroller-container:v1.18.0-8
+    resources:
+      requests:
+        cpu: 120m
+        memory: 100Mi
+      limits:
+        cpu: 250m
+        memory: 500Mi
   mlpipelineUI:
+    deploy: true
     image: quay.io/opendatahub/odh-ml-pipelines-frontend-container:beta-ui
+    resources:
+      limits:
+        cpu: 100m
+        memory: 256Mi
+      requests:
+        cpu: 100m
+        memory: 256Mi
     configMap: ds-pipeline-ui-configmap
   database:
-    mariaDB:   # mutually exclusive with custom
+    mariadb:   # mutually exclusive with custom
+      deploy: true
       image: registry.redhat.io/rhel8/mariadb-103:1-188
       username: mlpipeline
       pipelineDBName: randomDBName
+      pvcSize: 20Gi
+      resources:
+        requests:
+          cpu: 300m
+          memory: 800Mi
+        limits:
+          cpu: "1"
+          memory: 1Gi
       passwordSecret:
         name: ds-pipelines-db-sample
         key: password
-    # customDB:
-    #   host: mysql:3306
-    #   port: "8888"
-    #   username: root
-    #   pipelineDBName: randomDBName
-    #   passwordSecret:
-    #     name: somesecret
-    #     key: somekey
-  storage:
+#    externalDB:
+#      host: mysql:3306
+#      port: "8888"
+#      username: root
+#      pipelineDBName: randomDBName
+#      passwordSecret:
+#        name: somesecret
+#        key: somekey
+  objectStorage:
     minio:  # mutually exclusive with custom
+      deploy: true
       image: quay.io/opendatahub/minio:RELEASE.2019-08-14T20-37-41Z-license-compliance
       bucket: mlpipeline
+      resources:
+        requests:
+          cpu: 20m
+          memory: 100Mi
+        limits:
+          cpu: 250m
+          memory: 1Gi
       s3CredentialsSecret:
         secretName: somesecret-sample
         accessKey: accesskey
         secretKey: secretkey
-#    customStorage:
+#    externalStorage:
 #      host: minio.com
 #      port: "9092"
 #      bucket: mlpipeline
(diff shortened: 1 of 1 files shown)


HumairAK:
based on feedback from @rimolive: 

> **Ricardo Martinelli de Oliveira**
I think based on the proposal it sounds like each component is optional with the deploy: true . There might be a better way to do this.

> **Ricardo Martinelli de Oliveira**
I think if we're adding a DSP CR to an OpenShift cluster, all the minimum set of components should be deployed.

I suggest we can instead do the following as the minimal CR: 

```yaml
apiVersion: dspipelines.opendatahub.io/v1alpha1
kind: DSPipeline
metadata:
  name: sample
spec: {}
```

Where by default `deploy:` is available for all components, but is set as optional, and is by default set to `true`, this allows us to hide this field in non dev related contexts. 

In an empty spec, we assume mlpipeline-ui, mariadb, minio need to be provisioned for the user and are thus also deployed by DSPO.

gregsheremeta:
> I suggest we can instead do the following as the minimal CR:

+1, supporting optional deploy makes sense, and not littering the production CR with all those `deploy` lines is a nice way to achieve that

config/samples/dspipelines.io_v1alpha1_dspipeline.yaml
```diff
@@ -2,44 +2,105 @@ apiVersion: dspipelines.opendatahub.io/v1alpha1
...
-    # customDB:
-    #   host: mysql:3306
-    #   port: "8888"
-    #   username: root
-    #   pipelineDBName: randomDBName
-    #   passwordSecret:
-    #     name: somesecret
-    #     key: somekey
-  storage:
+#   customDB:
```
gmfrasca, line 84: does it make sense to have a `type: maria|custom|mysql|etc` field instead of making these mutex-able db items?

HumairAK: to me it makes sense to have custom (operator deploys a db for us) separate from maria|postgresql|mysql (we tell operator to use an existing db), since fields are different, and the handling operator logic is different (we supply our own db vs operator sets it up for us)

gregsheremeta: if the sub-fields are different, I think it's more typical to have separate objects

an example that jumped to mind is the openshift install config where you set platform-specific stuff:

```yaml
compute:
- name: worker
  platform:
    aws:
      amiID: ami-06c4d345f7c207239 
      type: m5.4xlarge
```

aws, gcp, etc are all similar but slightly different fields, i.e. `amiID` means nothing in gcp

so gcp has its own object that doesn't include an `amiID`

config/samples/dspipelines.io_v1alpha1_dspipeline.yaml
```diff
@@ -2,44 +2,105 @@ apiVersion: dspipelines.opendatahub.io/v1alpha1
...
   apiServer:
+    deploy: true
     # optional (default is: ds-pipeline-artifact-script-${metadata.name} / artifact_script)
     artifactScriptConfigMap:
       name: ds-pipeline-artifact-script-sample
       key: "artifact_script"
     apiServerImage: quay.io/modh/odh-ml-pipelines-api-server-container:v1.18.0-8
     artifactImage: quay.io/modh/odh-ml-pipelines-artifact-manager-container:v1.18.0-8
+    dBConfigConMaxLifetimeSec: 120
+    CollectMetricsFlag: true
```
gregsheremeta, line 16: what's the significance of this being PascalCase vs camelCase? (ref: https://github.com/kubeflow/kfp-tekton/blob/master/backend/src/apiserver/main.go#L52)

gregsheremeta:
other than my silly question about the `C`, I don't have anything to add. Happy to tag this as-is if there are no other outstanding comments.

config/samples/dspipelines.io_v1alpha1_dspipeline.yaml
HumairAK: this should be camelcase so it's consistent with the rest of the spec

HumairAK:
Since `.status` is also part of the API Spec, we should also make that part of the discussion. Kubernetes community offers standard status properties for controllers to employ known as _Conditions_. They outline best practices and conventions for this [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties). 

Based on my understanding, we should have at the minimum the following: 

```yaml

status: 
  conditions:
    - type: Ready
      status: True
      observedGeneration: 4
      lastTransitionTime: '2023-02-02T21:00:00Z'
      reason: MinimumReplicasAvailable  # other meaningful one word reasons go here
      message: 'some message'
```
This is consistent with the api conventions from the k8s community, as mentioned in the link above: 

```
When designing Conditions for a resource, it's helpful to have a common top-level condition which summarizes more detailed conditions. Simple consumers may simply query the top-level condition. 
```

Now DSPO deploys different components, and we would need a concrete definition of what it means for a `DSPipeline` to be `Ready`. A more beefier `.status` spec could be: 

<details> 

<summary> detailed_status_spec.yaml </summary>

```yaml
status:
  # Reports True iff:
  # * apiServerConditions, databaseConditions, objectStorageConditions, persistenceAgentConditions report (Ready : True)
  # AND
  # * mLPIpelinesUIConditions is (Ready: True) OR is (Ready: False && DeploymentDisabled)
  conditions:
    - type: Ready
      status: True
      observedGeneration: 4
      lastTransitionTime: '2023-02-02T21:00:00Z'
      reason: MinimumReplicasAvailable
      message: 'some message'
  componentsConditions:
    apiServerConditions:
      - type: Ready
        status: True
        observedGeneration: 4
        lastTransitionTime: '2023-02-02T21:00:00Z'
        reason: MinimumReplicasAvailable
        message: 'some message'
    mLPIpelinesUIConditions:
      - status: True
        type: Ready
        observedGeneration: 4
        lastTransitionTime: '2023-02-02T21:00:00Z'
        reason: MinimumReplicasAvailable # DeploymentDisabled
        message: 'some message'
    persistenceAgentConditions:
      - status: True
        type: Ready
        observedGeneration: 4
        lastTransitionTime: '2023-02-02T21:00:00Z'
        reason: MinimumReplicasAvailable
        message: 'some message'
    scheduledWorkflowConditions:
      - status: True
        type: Ready
        observedGeneration: 4
        lastTransitionTime: '2023-02-02T21:00:00Z'
        reason: MinimumReplicasAvailable
        message: 'some message'
    # Do we need to do this?? API Server application already
    # checks for db/storage connectivity, and pod will fail to come up
    # in such a case.
    databaseConditions:
      - status: False
        type: Ready
        observedGeneration: 4
        lastTransitionTime: '2023-02-02T21:00:00Z'
        reason: DataBaseUnreachable # DataBaseFailedToDeploy, DataBaseReady
        message: '500 gateway error received'
    objectStorageConditions:
      - status: False
        type: Ready
        observedGeneration: 4
        lastTransitionTime: '2023-02-02T21:00:00Z'
        reason: ObjectStorageUnreachable  # ObjectStorageFailedToDeploy, ObjectStorageReady
        message: 'host unreachable'

```


</details>

We don't have to go with something this detailed, but if we go with a single top level condition, we should have some explicit definition of what it means to be `Ready`.

HumairAK:
Deleted comments were for CI testing and unrelated to this discussion, please ignore.

gregsheremeta:
I don't think I've personally seen nested conditions (componentsConditions.apiServerConditions) -- more used to a giant flat list (top level Ready, then ApiServerReady maybe?), but I'm ok with nesting if that's allowed.

Top level Ready and subcomponent Ready makes sense. We can also add more granular conditions as we discover use cases down the line.

> Do we need to do this?? API Server application already checks for db/storage connectivity, and pod will fail to come up
in such a case

Seems useful to me. Covers the case where the API comes up and the db dies later. In that case, the top level Ready could report something generic like "database not ready" and the database condition could report specifics

HumairAK:
> I don't think I've personally seen nested conditions (componentsConditions.apiServerConditions) -- more used to a giant flat list (top level Ready, then ApiServerReady maybe?), but I'm ok with nesting if that's allowed.

I think I like the idea of flattening it out more, will update

HumairAK:
updated status: 


<details>

<summary>detailed_status_spec.yaml</summary>

```yaml
status:
  conditions:
    - type: Ready
      status: True
      observedGeneration: 4
      lastTransitionTime: '2023-02-02T21:00:00Z'
      reason: MinimumReplicasAvailable
      message: 'some message'
    - type: ApiServerReady
      status: True
      observedGeneration: 4
      lastTransitionTime: '2023-02-02T21:00:00Z'
      reason: MinimumReplicasAvailable
      message: 'some message'
    - status: True
      type: MLPIpelinesUIReady
      observedGeneration: 4
      lastTransitionTime: '2023-02-02T21:00:00Z'
      reason: MinimumReplicasAvailable # DeploymentDisabled
      message: 'some message'
    - status: True
      type: PersistenceAgentReady
      observedGeneration: 4
      lastTransitionTime: '2023-02-02T21:00:00Z'
      reason: MinimumReplicasAvailable
      message: 'some message'
    - status: True
      type: ScheduledWorkflowReady
      observedGeneration: 4
      lastTransitionTime: '2023-02-02T21:00:00Z'
      reason: MinimumReplicasAvailable
      message: 'some message'
    - status: False
      type: DatabaseReady
      observedGeneration: 4
      lastTransitionTime: '2023-02-02T21:00:00Z'
      reason: DataBaseUnreachable # DataBaseFailedToDeploy, DataBaseReady
      message: '500 gateway error received'
    - status: False
      type: ObjectStorageReady
      observedGeneration: 4
      lastTransitionTime: '2023-02-02T21:00:00Z'
      reason: ObjectStorageUnreachable  # ObjectStorageFailedToDeploy, ObjectStorageReady
      message: 'host unreachable'

```


</details>

config/samples/dspipelines.io_v1alpha1_dspipeline.yaml
HumairAK: As per in call discussion, we have settled for having external db connection as a separate dict than mariadb deployment dict. I've updated the cr to reflect this change.

PR #20, opened 2023-02-16
Improve custom artifact script handling.

### LINKED ISSUE #19: Add better handling of custom artifact script
Currently the custom artifact script that a user can provide for the api server is not handled in any meaningful way. 

1. The name of the default script is hardcoded with a suffix "-sample", this should not be the case, the suffix should always be the metadata.name for the complementing DSPipeline CR 
2. We should fail to deploy api server if configmap that is provided by end user does not exist, (this error should be delegated to the `.status` field of the DSPipeline CR but that is out of scope of this issue).

HumairAK:
This change will make it so api server will fail to deploy if a custom artifact script provided by the user does not exist. Furthermore, it will also ensure that the default artifact script follows naming conventions that are consistent with the rest of the namespace scoped resources. Tests are addded to accomodate this change.

## Description
Resolve #19 

## How Has This Been Tested?
Via unit tests, and live ocp cluster.

## Merge criteria:

### DIFF
--- a/config/samples/dspipelines.io_v1alpha1_dspipeline.yaml
+++ b/config/samples/dspipelines.io_v1alpha1_dspipeline.yaml
@@ -2,110 +2,52 @@ apiVersion: dspipelines.opendatahub.io/v1alpha1
 kind: DSPipeline
 metadata:
   name: sample
-  namespace: data-science-project
 spec:
   apiServer:
-    deploy: true
-    image: quay.io/modh/odh-ml-pipelines-api-server-container:v1.18.0-8
-    resources:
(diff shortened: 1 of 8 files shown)


HumairAK:
[This pr](https://github.com/opendatahub-io/data-science-pipelines-operator/pull/8) did not have the code changes to go along with it. This was resulting in a conflict with this pr which is applying a change to the original cr spec. 


I've added another commit that keeps the old cr spec as the standard until another follow up pr to #8 with the code changes takes effect. The proposed cr spec is also kept but the file is named as such.

PR #22, opened 2023-02-20
Adding a service to expose Operator SDK metrics

DharmitD:
## Description

Adding a service to expose Operator SDK metrics port

## How Has This Been Tested?

On a test cluster:

- Added this service config
- Port forwarded to the localhost port via the service:
`oc port-forward svc/data-science-pipelines-operator 8080:8080 -n data-science-pipelines-operator`
- Curled to the localhost to retrieve a list of Operator SDK metrics:
`curl -v http://localhost:8080/metrics`

Find a list of metrics returned by the curl command [here.](http://pastebin.test.redhat.com/1092224)

## Merge criteria:

### DIFF
--- a/config/manager/manager-service.yaml
+++ b/config/manager/manager-service.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: data-science-pipelines-operator
+  labels:
+    control-plane: controller-manager
+spec:
+  ports:
+    - name: metrics
+      port: 8080
+  selector:
+    control-plane: controller-manager
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -1,2 +1,3 @@
 resources:
 - manager.yaml
+- manager-service.yaml


gregsheremeta:
content looks good to me. The only question I have is whether we'd separate out the service into its own config file. The other config directories seem to have resources be mostly separate (for example: https://github.com/opendatahub-io/data-science-pipelines-operator/tree/main/config/rbac), and I'm not familiar enough with the conventions. Perhaps another reviewer can make a suggestion.

config/manager/manager-service.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: data-science-pipelines-operator-controller-manager
```
HumairAK, line 4: can you shorten the name to `data-science-pipelines-operator`

HumairAK (changes requested):

HumairAK:
Can you confirm that the servicemonitor works? Confirm this by checking if the metrics show up in ocp monitoring stack. My guess is it won't work because it's missing the labels for the selector here: https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/config/prometheus/monitor.yaml 

The `port` will also need to be updated to match the exposed port in the `service`.

gregsheremeta:
> My guess is it won't work because it's missing the labels for the selector here

By "it", did you mean the service? The service has to have the same labels as the servicemonitor?

HumairAK:
> By "it", did you mean the service? The service has to have the same labels as the servicemonitor? 

yeah the service monitor will select on the labels for the service

HumairAK:
@DharmitD you should also test by port-forwarding the service, not the pod `oc port-forward service/$service_name` since that is the change you are making

config/manager/manager-service.yaml
```diff
@@ -0,0 +1,10 @@
...
+spec:
+  ports:
+    - name: http
```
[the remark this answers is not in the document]
DharmitD: Done, updated the port name.

config/manager/manager-service.yaml
DharmitD: Done, updated the service name.

DharmitD:
> @DharmitD you should also test by port-forwarding the service, not the pod `oc port-forward service/$service_name` since that is the change you are making

I updated the service to add the same label as the service monitor. Tried port forwarding through the service:
`oc port-forward svc/data-science-pipelines-operator 8080:8080 -n data-science-pipelines-operator`

The port forward worked, further queried metrics using `curl`:
`curl http://localhost:8080/metrics`
Curl returned the same set of metrics.

DharmitD:
> Can you confirm that the servicemonitor works? Confirm this by checking if the metrics show up in ocp monitoring stack. My guess is it won't work because it's missing the labels for the selector here: https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/config/prometheus/monitor.yaml
> 
> The `port` will also need to be updated to match the exposed port in the `service`.

Added a ServiceMonitor config in https://github.com/opendatahub-io/data-science-pipelines-operator/pull/24 with labels matching to the above service config.
The ServiceMonitor works and metrics can be queried.

PR #27, opened 2023-02-27
Cr proposed spec logic

HumairAK:
This PR adds code logic to apply the proposal changes accepted [here](https://github.com/opendatahub-io/data-science-pipelines-operator/pull/8).

The PR is a bit big, it's broken in multiple commits, I will summarize the changes here with detailed instructions on how to test this PR

## Description
Major changes are as follows: 

* API has been updated to include new CR spec fields (location: `api/v1alpha1*`)
* CRD has been updated to match updated api spec (location: `config/crd/`)
* Parameter logic has been updated and simplified to source directly from CR (location: `controllers/dspipeline_params.go`)
* Manifest templates have been updated to match new parameter fields (basically all the changes in `config/internal`)
* sample manifests have been updated to provide a standard dsp cr, simplified (empty) cr, and a cr with custom configs (readme updated on how to provide your own creds)
* added logic to enable `deploy` functionality for each component
* added logic to handle custom secrets as identified in the spec proposal (location: `controllers/dspipeline_params.go`)
* disabled viewercrd-controllers' default deployment (if you provide empty `spec` it won't be deployed)
* unit tests for testing the cases found in `controllers/testdata/deploy` see `controllers/dspipeline_controller_test.go` for their description.

## How Has This Been Tested?
This PR has been tested locally and on a live ocp cluster. Testing this PR is relatively easy, reviewer should attempt the following: 

```bash
git clone $REPO
cd data-science-pipelines-operator
git checkout $PR-Branch
```

**Run Unit tests**

Follow instructions [here](https://github.com/opendatahub-io/data-science-pipelines-operator#run-tests) to run unit tests.

The tests deploy the custom resources in `controllers/testdata/deploy`, and checks whether the expected resources are created by matching with manifests in `controllers/testdata/results`. 

**Build & Deploy the PR**
You can build your own image or use the one [here](https://quay.io/hukhan/data-science-pipelines-operator:v0.0.7): 

Ensure you `oc login` to a dev OCP cluster, have OCP pipelines operator installed. 

```bash
```

**Deploy DSP**

Once DSPO is running, try creating an empty CR as provided in `config/samples/dspipelines_simple.yaml`. 

Once you submit it, inspect the resulting CR filled in with defaults. Wait for kfp to deploy, ensure all pods are created as expected and you can create a `run`. 

In a different namespace try to deploy DSP with custom configs, as specified in the updated README.md with the manifests found in `config/samples/custom-configs/dspipelines.io_v1alpha1_dspipeline.yaml`. 

You can investigate these secrets for the credentials and see if you can access MariaDB and Minio using these creds.

For example for MariaDB the creds are: 

```yaml
stringData:
  password: testingpassword
```

Ssh into the mariaDB pod and run the following: 

```pbash
mysql -u mlpipeline -h 0.0.0.0 --port=3306 --password="testingpassword"
use test;
CREATE TABLE example (node_id INT PRIMARY KEY, node_name VARCHAR(30));
INSERT INTO test.example VALUES (1, 'someentry');
SELECT * FROM test.example;
```

You can then try changing the secret creds, restart the pod, and run `SELECT * FROM test.example;` and confirm the data is still there, but the creds are updated. You can do a similar test with Minio.

## Merge criteria:


HumairAK:
Note that the CRD Name change from `DSPipeline` to `DataSciencePipelineAplication` is not included in this PR. This PR was already getting too big and such a change would explode the amount of changes, and as it's mostly refactoring, I think it's better to include that as a separate Issue/PR.

DharmitD:
Followed steps listed to test out the CR creation as listed.
1. Unit tests went through, with a coverage of 77.8% reported.
2. Able to build and deploy PR changes, DSP Operator comes up successfully.
3. When applying the CR `config/samples/custom-configs/dspipelines_simple.yaml` to deploy a DSP instance, relevant resources don't come up. On investigating the DSPO leader pod logs further, I see this error message:
```
2023-03-01T18:12:15Z	ERROR	Reconciler error	{"controller": "dspipeline", "controllerGroup": "dspipelines.opendatahub.io", "controllerKind": "DSPipeline", "DSPipeline": {"name":"sample","namespace":"test-dspo1"}, "namespace": "test-dspo1", "name": "sample", "reconcileID": "df99cf83-924a-41bf-9d00-9f869aea4afe", "error": "Secret \"testdbsecret\" not found"}
```

@gregsheremeta and I tried deploying with removing the secret ref instance, and we were able to stand up DSP successfully.

HumairAK:
@DharmitD 

> When applying the CR config/samples/custom-configs/dspipelines_simple.yaml to deploy a DSP instance, relevant resources don't come up. On investigating the DSPO leader pod logs further, I see this error message:

this is expected behavior because you only deployed the CR and not all the configmaps/secrets in that directory, which is what that example is showcasing, do a kustomize build and oc apply all the resulting manifests

DharmitD:
> @DharmitD
> 
> > When applying the CR config/samples/custom-configs/dspipelines_simple.yaml to deploy a DSP instance, relevant resources don't come up. On investigating the DSPO leader pod logs further, I see this error message:
> 
> this is expected behavior because you only deployed the CR and not all the configmaps/secrets in that directory, which is what that example is showcasing, try doing a kustomize build on the custom-config folder and oc apply all the resulting manifests

Done, did a kustomize build in the `custom-config` directory and and applied manifests, a DSP instance was created as expected.

Further was able to query the database successfully too:
```
MariaDB [test]> SELECT * FROM test.example;
+---------+-----------+
| node_id | node_name |
+---------+-----------+
|       1 | someentry |
+---------+-----------+
1 row in set (0.000 sec)
```

DharmitD (commented):
/lgtm

Deployed and tested as per instructions listed, could deploy DSPO and stand up a DSP instance.

controllers/testutil/util.go
```diff
@@ -40,60 +40,60 @@ func ConfigMapsAreEqual(expected v1.ConfigMap, actual v1.ConfigMap) bool {
...
 		return false
 	}
-	for i := range dp1.Spec.Template.Spec.Containers {
-		c1 := dp1.Spec.Template.Spec.Containers[i]
-		c2 := dp2.Spec.Template.Spec.Containers[i]
-		if !reflect.DeepEqual(c1.Env, c2.Env) {
+	for i := range expectedDep.Spec.Template.Spec.Containers {
+		expectedContainer := expectedDep.Spec.Template.Spec.Containers[i]
+		actualContainer := actualDep.Spec.Template.Spec.Containers[i]
+		if !reflect.DeepEqual(expectedContainer.Env, actualContainer.Env) {
```
gmfrasca, line 72: this (essentially) case block feels pretty inelastic, it would be nice if we could define a list of attributes and dynamically iterate through them

config/samples/custom-configs/artifact_script.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1
+data:
+  somekey: |-
```
gmfrasca, line 3: we should name this something better than `somekey`

controllers/testutil/util.go
HumairAK: yeah I agree, the problem is we do not want to compare all fields (some fields will always be different, e.g. `creationtimestamp`, and there will be some opinionated stances on what ought to be compared, which is probably why an equals operation isn't provided for us already by client-go. I'm open to suggestions though.

HumairAK: One suggestion I've seen is to hash a resource and store it in an annotation before submitting it to the api server, so you can always retrieve it's "vanilla" state before the cluster adds it's all cluster specific bloat. Then you can retrieve it and compare it with another hashed resource. 

I wasn't a huge fan of this method though.

gmfrasca:
should the test case nomiclature in `testdata` be updated to reflect what each case is actually testing?  just naming them `case_0`, `case_1`, etc isn't very helpful

PR #31, opened 2023-03-03
Parameterize images

HumairAK:
## Description
This change introduces the ability to pass in default images via either a configmap or env vars. Environment variables take precedence over configmaps. Environment variables follow the same nested behavior as the yaml config equivalent. For example "images.apiserver" in a `config.yaml` passed in to the operator is the equivalent to passing in an ENV Variable `IMAGES_APISERVER`.

## How Has This Been Tested?
Locally and on OCP OSD with: quay.io/hukhan/data-science-pipelines-operator:v0.0.9
Run unit tests on this pr by cloning the branch and following instructions [here](https://github.com/opendatahub-io/data-science-pipelines-operator#run-tests). 

Deploy on dev cluster with ocp pipelines (logged in as cluster admin): 

```
```

Once the Operator is deployed create a DSPA cr: 

Deploy a sample cr with the updated resource kind

```yaml
metadata:
  name: sample
spec: {}
```

Inspect the images, now go to the DSPO deployment and change the Image ENV vars, and confirm the change is delegated to the DSPA component that was updated. 

Remove the Environment variables, and confirm the configmap passed in values are consumed instead. Test this by updating the configmap after env vars are removed, and restarting the deployment.

## Merge criteria:


HumairAK:
See [this change](https://github.com/opendatahub-io/data-science-pipelines-operator/commit/1a74e6ff4d1dec15a6c4e310090db97ac1948c70) for how new image versions will need to be updated given this change.

HumairAK:
With this change we can then paramerize the env vars as kfdef params, and sub them in via kfdef similar to [this](https://github.com/red-hat-data-services/odh-deployer/blob/main/kfdefs/rhods-model-mesh.yaml#L9-L22).

HumairAK:
@DharmitD w.r.t  to [this comment ](https://github.com/opendatahub-io/data-science-pipelines-operator/pull/32#issuecomment-1453647772)

> Would we need to update these as well? 

No, tests have their own default configs that are independent of what's deployed in live environment by default, see chagnes in `controllers/testdata/deploy` for these configs

harshad16 (commented):
/lgtm

great work

config/configmaps/files/config.yaml
```diff
@@ -0,0 +1,11 @@
+images:
+  apiServer: quay.io/modh/odh-ml-pipelines-api-server-container:v1.18.0-8
+  artifact: quay.io/modh/odh-ml-pipelines-artifact-manager-container:v1.18.0-8
+  persistentAgent: quay.io/modh/odh-ml-pipelines-persistenceagent-container:v1.18.0-8
+  scheduledWorkflow: quay.io/modh/odh-ml-pipelines-scheduledworkflow-container:v1.18.0-8
+  viewerCRD: quay.io/modh/odh-ml-pipelines-viewercontroller-container:v1.18.0-8
+  cache: registry.access.redhat.com/ubi8/ubi-minimal
+  moveResultsImage: busybox
```
gmfrasca, line 8: we should at a minimum mirror `busybox` to a quay repo to avoid rate limits

config/manager/manager.yaml
```diff
@@ -24,13 +24,40 @@ spec:
...
+          - name: IMAGES_PERSISTENTAGENT
+            value: quay.io/modh/odh-ml-pipelines-persistenceagent-container:v1.18.0-8
+          - name: IMAGES_SCHEDULEDWORKFLOW
+            value: quay.io/modh/odh-ml-pipelines-scheduledworkflow-container:v1.18.0-8
+          - name: IMAGES_VIEWERCRD
+            value: quay.io/modh/odh-ml-pipelines-viewercontroller-container:v1.18.0-8
+          - name: IMAGES_CACHE
+            value: registry.access.redhat.com/ubi8/ubi-minimal
+          - name: IMAGES_MOVERESULTSIMAGE
+            value: busybox
```
gmfrasca, line 54: see comment above

controllers/testdata/deploy/case_0/config.yaml
```diff
@@ -0,0 +1,11 @@
+images:
+  apiServer: quay.io/modh/odh-ml-pipelines-api-server-container:v1.18.0-8
+  artifact: quay.io/modh/odh-ml-pipelines-artifact-manager-container:v1.18.0-8
+  persistentAgent: quay.io/modh/odh-ml-pipelines-persistenceagent-container:v1.18.0-8
+  scheduledWorkflow: quay.io/modh/odh-ml-pipelines-scheduledworkflow-container:v1.18.0-8
+  viewerCRD: quay.io/modh/odh-ml-pipelines-viewercontroller-container:v1.18.0-8
+  cache: registry.access.redhat.com/ubi8/ubi-minimal
+  moveResultsImage: busybox
```
gmfrasca, line 8: quay

controllers/testdata/deploy/case_2/config.yaml
```diff
@@ -0,0 +1,11 @@
+images:
+  apiServer: quay.io/modh/test-api-server-container:test
+  artifact: quay.io/modh/test-artifact-manager-container:test
+  persistentAgent: quay.io/modh/test-persistenceagent-container:test
+  scheduledWorkflow: quay.io/modh/test-scheduledworkflow-container:test
+  viewerCRD: quay.io/modh/test-viewercontroller-container:test
+  cache: registry.access.redhat.com/ubi8/test-ubi-minimal
+  moveResultsImage: testbusybox
```
gmfrasca, line 8: quay. also, is this an actual thing?

controllers/config/defaults.go
```diff
@@ -56,8 +49,41 @@ const (
...
+	APIServerImagePath            = "images.apiServer"
+	APIServerArtifactImagePath    = "images.artifact"
+	PersistenceAgentImagePath     = "images.persistentAgent"
+	ScheduledWorkflowImagePath    = "images.scheduledWorkflow"
+	ViewerCRDImagePath            = "images.viewerCRD"
+	APIServerCacheImagePath       = "images.cache"
+	APIServerMoveResultsImagePath = "images.moveResultsImage"
+	MlPipelineUIImagePath         = "images.mlPipelineUI"
+	MariaDBImagePath              = "images.mariaDB"
+	MinioImagePath                = "images.minio"
```
gmfrasca, line 63: I think the convention is to use PascalCase for viper configs, right?

controllers/dspipeline_controller_test.go
```diff
@@ -130,12 +130,12 @@ func deleteDSP(path string, opts mf.Option) {
 	dsp2 := &dspav1alpha1.DataSciencePipelinesApplication{}
 	Eventually(func() error {
 		namespacedNamed := types.NamespacedName{Name: dsp.Name, Namespace: WorkingNamespace}
-		Expect(k8sClient.Get(ctx, namespacedNamed, dsp2)).NotTo(HaveOccurred())
-		if !reflect.DeepEqual(dsp2, &dspav1alpha1.DataSciencePipelinesApplication{}) {
-			return errors.New("DSP still exists on cluster")
+		err = k8sClient.Get(ctx, namespacedNamed, dsp2)
+		if err != nil && apierrs.IsNotFound(err) {
```
gmfrasca, line 134: this looks like we're doing some sort of 'unknown error' ignoring here.  Is that the case, and if so why?

config/configmaps/files/config.yaml
HumairAK: this is not scope in this pr, addressed in this PR: https://github.com/opendatahub-io/data-science-pipelines-operator/pull/32/files

config/manager/manager.yaml
HumairAK: this is not scope in this pr, addressed in this PR: https://github.com/opendatahub-io/data-science-pipelines-operator/commit/32bb5f78cd24dcb51e63c370c6ad054413446f32

controllers/testdata/deploy/case_1/config.yaml
```diff
@@ -0,0 +1,11 @@
+images:
+  apiServer: quay.io/modh/odh-ml-pipelines-api-server-container:v1.18.0-8
+  artifact: quay.io/modh/odh-ml-pipelines-artifact-manager-container:v1.18.0-8
+  persistentAgent: quay.io/modh/odh-ml-pipelines-persistenceagent-container:v1.18.0-8
+  scheduledWorkflow: quay.io/modh/odh-ml-pipelines-scheduledworkflow-container:v1.18.0-8
+  viewerCRD: quay.io/modh/odh-ml-pipelines-viewercontroller-container:v1.18.0-8
+  cache: registry.access.redhat.com/ubi8/ubi-minimal
+  moveResultsImage: busybox
```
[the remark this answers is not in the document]
HumairAK: see above

controllers/testdata/deploy/case_2/config.yaml
HumairAK: this is a a funcitonal test (as denoted by its location), it doesn't matter if this exists, it's not actually used to deploy thing, we're just testing to see if the image config gets repalced

controllers/testdata/deploy/case_3/config.yaml
```diff
@@ -0,0 +1,11 @@
+images:
+  apiServer: quay.io/modh/test-api-server-container:test1
+  artifact: quay.io/modh/test-artifact-manager-container:test1
+  persistentAgent: quay.io/modh/test-persistenceagent-container:test1
+  scheduledWorkflow: quay.io/modh/test-scheduledworkflow-container:test1
+  viewerCRD: quay.io/modh/test-viewercontroller-container:test1
+  cache: registry.access.redhat.com/ubi8/test-ubi-minimal
+  moveResultsImage: testbusybox
```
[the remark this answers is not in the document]
HumairAK: see above

controllers/dspipeline_controller_test.go
HumairAK: Can you explain? 

This makes sense when the whole code block is viewed: 

```golang
	Eventually(func() error {
		namespacedNamed := types.NamespacedName{Name: dsp.Name, Namespace: WorkingNamespace}
		err = k8sClient.Get(ctx, namespacedNamed, dsp2)
		if err != nil && apierrs.IsNotFound(err) {
			return nil
		}
		return fmt.Errorf("resource still exists on cluster")
	}, timeout, interval).ShouldNot(HaveOccurred())
```

This tests an async call, which is basically doing: 

Try to fetch this DSP resource, **we are expecting** it to not be on the cluster (IsNotFound). **We are not expecting** it to exist on the cluster, or encounter any other error while attempting to fetch.

controllers/config/defaults.go
HumairAK: I'm not sure why Viper would impose a convention on how applications handle their config files, afaik the convention for yaml files is to do do camelcase and that's what I tried to adhere to.

controllers/dspipeline_controller_test.go
HumairAK: I see what you are saying, there's a missing case here, I'll update

HumairAK: updated

controllers/config/defaults.go
HumairAK: I updated it to pascal case regardless, this is what dspipelines app does it, might as well stay consistent with that.

gmfrasca (approved):

controllers/dspipeline_params.go
```diff
@@ -124,7 +124,7 @@ func (p *DSPAParams) SetupDBParams(ctx context.Context, dsp *dspa.DataSciencePip
 		if p.MariaDB == nil {
 			p.MariaDB = &dspa.MariaDB{
 				Deploy:    true,
-				Image:     config.MariaDBImage,
+				Image:     config.GetStringConfigWithDefault(config.MariaDBImagePath, config.DefaultImageValue),
```
gregsheremeta, line 127: @HumairAK the only thing I didn't understand about this PR is this `GetStringConfigWithDefault`. It looks like `DefaultImageValue` is a const set to `MustSetInConfig`, so there's not really any useful fallback value ever returned from `GetStringConfigWithDefault`. What is your intent there? Is it to make sure image pulls fail with `cannot pull docker/MustSetInConfig` ? Is that even what would happen?

Perhaps a comment on `GetStringConfigWithDefault` would help.

HumairAK: @gregsheremeta you're right the default here will never be used, it was mostly to just stay consistent with the rest of the pattern and add additional future proof protection to the logic. 

as you already know, this would never happen because we enforce the[ requirement of the Image config](https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/main.go#L80-L84). 

however to me this seemed a bit too decoupled from the rest of the params logic (relying on configs/env vars), so it made sense to have this additional check. Otherwise we are relying on the config file or env variables to be set correctly, are not empty, etc.

Having said that, I'm not against just setting the image directly without using `GetStringConfigWithDefault`.

HumairAK: > which pattern do you mean? 

Maybe "rest of the code" would have been more appropriate, essentially the checks we perform on other parameters.  

> what's an example of a future change that this is protecting against? 

* config is made optional/removed
* env vars / configs are empty (accidentally or by design) 

> that makes sense, but I think you could also just panic about it with a direct message instead of causing an image pull error in a roundabout way

We would never get a pull error since this this should never happen any way. 

Maybe a better way to go about this is to just put in a test that checks for all these things, since we're already accounting for it in config logic, I don't mind either way ¯\_(ツ)_/¯

Issue #9, opened 2023-02-10
Update auth delegator CRB generation logic


anishasthana:
Currently, the auth delegator cluster rolebinding that is created is of the format `<standard-prefix>+<custom_resource_name>`

This can result in CRB conflicts if a custom resource with the same name is created in multiple namespaces. I'd recommend changing the CRB name to also include the namespace name, so something like  `<standard-prefix>+<namespace>+<custom_resource_name>`

PR #34, opened 2023-03-03
Add enable route functionality.

HumairAK:
## Description

This change introduces the option for the CR submitter to enable/disable routes for the Apiserver. When the route is enabled, an oauth proxy container is added to the api server and a k8s route is created.

To ensure that all external traffic goes through the route, and all pod-to-pod communications to apiserver go through port 8888 (default), we use a k8s network policy to enforce this behavior.

By allowing pods to communicate to apiserver via port 8888, we can continue provisioning mlpipelines-ui alongside this feature, otherwise the ui would also need to go through oauth proxy.

If the CR submitter switches enableRoute from 'true' to 'false' then the operator will clean up the existing route on the cluster.

## How Has This Been Tested?
Locally and on OCP OSD with: quay.io/hukhan/data-science-pipelines-operator:pull-33
Run unit tests on this pr by cloning the branch and following instructions [here](https://github.com/opendatahub-io/data-science-pipelines-operator#run-tests). 

Deploy on dev cluster with ocp pipelines (logged in as cluster admin): 

```
```

Deploy a sample cr with the updated resource kind: 

```yaml
metadata:
  name: sample
spec: {}
```

By default the field `enableRoute` will be set to `true`.  

Login to the UI,  confirm you can login w/o issues, upload/start a run.

Navigate to the Routes page and fetch the apiserver route, or run: 
```bash
# assuming namespace: data-science-project, and cr name: sample
ROUTE=`https://$(oc get routes -n data-science-project ds-pipeline-sample --template={{.spec.host}})`
```

Curl this route and see if you can by pass the oauth proxy and fetch runs: 

```bash
curl --insecure --location --request GET "${ROUTE}/apis/v1beta1/runs"  --header "Authorization: Bearer $(oc whoami --show-token)"
```

Further testing can include confirming the network policy works. Try curling the service for the api server from another pod that is not deployed by dspo: 

This should fail: 

```
$ oc -n openshift-monitoring rsh alertmanager-main-0
$ curl --location --request GET http://ds-pipeline-sample.data-science-project.svc.cluster.local:8888/apis/v1beta1/runs
```
The above command will succeed if you try it from a dspo component pod, for example the associated persistenceagent pod.

This should succeed: 

```bash
# YOUR_TOKEN == $(oc whoami --show-token)
curl --location --request GET https://ds-pipeline-sample.data-science-project.cluster.local:8443/apis/v1beta1/runs --header 'Authorization: Bearer ${YOUR_TOKEN}' --insecure
```

This also resolves: https://github.com/opendatahub-io/data-science-pipelines-operator/issues/9

## Merge criteria:

### DIFF
--- a/api/v1alpha1/dspipeline_types.go
+++ b/api/v1alpha1/dspipeline_types.go
@@ -71,6 +71,9 @@ type APIServer struct {
 	CollectMetrics bool `json:"collectMetrics"`
 	// +kubebuilder:default:=true
 	// +optional
+	EnableRoute bool `json:"enableOauth"`
+	// +kubebuilder:default:=true
+	// +optional
 	AutoUpdatePipelineDefaultVersion bool                  `json:"autoUpdatePipelineDefaultVersion"`
 	Resources                        *ResourceRequirements `json:"resources,omitempty"`
 }
--- a/config/internal/apiserver/deployment.yaml.tmpl
+++ b/config/internal/apiserver/deployment.yaml.tmpl
@@ -18,6 +18,54 @@ spec:
         component: data-science-pipelines
     spec:
       containers:
+        {{ if .APIServer.EnableRoute }}
+        - name: oauth-proxy
+          args:
+            - --https-address=:8443
+            - --provider=openshift
+            - --openshift-service-account=ds-pipeline-{{.Name}}
+            - --upstream=http://localhost:8888
+            - --tls-cert=/etc/tls/private/tls.crt
+            - --tls-key=/etc/tls/private/tls.key
+            - --cookie-secret=SECRET
+            - '--openshift-delegate-urls={"/": {"group":"route.openshift.io","resource":"routes","verb":"get","name":"ds-pipeline-{{.Name}}","namespace":"{{.Namespace}}"}}'
+            - '--openshift-sar={"namespace":"{{.Namespace}}","resource":"routes","resourceName":"ds-pipeline-{{.Name}}","verb":"get","resourceAPIGroup":"route.openshift.io"}'
+            - --skip-auth-regex='(^/metrics|^/apis/v1beta1/healthz)'
+          image: registry.redhat.io/openshift4/ose-oauth-proxy:v4.12.0
+          ports:
+            - containerPort: 8443
+              name: oauth
+          livenessProbe:
+            httpGet:
+              path: /oauth/healthz
+              port: oauth
+              scheme: HTTPS
+            initialDelaySeconds: 30
+            timeoutSeconds: 1
+            periodSeconds: 5
+            successThreshold: 1
+            failureThreshold: 3
+          readinessProbe:
+            httpGet:
+              path: /oauth/healthz
+              port: oauth
+              scheme: HTTPS
+            initialDelaySeconds: 5
+            timeoutSeconds: 1
+            periodSeconds: 5
+            successThreshold: 1
+            failureThreshold: 3
+          resources:
+            limits:
+              cpu: 100m
+              memory: 256Mi
+            requests:
+              cpu: 100m
+              memory: 256Mi
+          volumeMounts:
+            - mountPath: /etc/tls/private
+              name: proxy-tls
+        {{ end }}
         - env:
             - name: POD_NAMESPACE
               value: "{{.Namespace}}"
@@ -143,3 +191,7 @@ spec:
               {{ end }}
             {{ end }}
       serviceAccountName: ds-pipeline-{{.Name}}
+      volumes:
+        - name: proxy-tls
+          secret:
+            secretName: ds-pipelines-proxy-tls-{{.Name}}
--- a/config/internal/apiserver/route.yaml.tmpl
+++ b/config/internal/apiserver/route.yaml.tmpl
@@ -1,0 +1,19 @@
+kind: Route
+apiVersion: route.openshift.io/v1
+metadata:
+  name: ds-pipeline-{{.Name}}
+  namespace: {{.Namespace}}
+  labels:
+    app: ds-pipeline-{{.Name}}
+    component: data-science-pipelines
+  annotations:
+    kubernetes.io/tls-acme: "true"
+spec:
+  to:
+    kind: Service
+    name: ds-pipeline-{{.Name}}
+    weight: 100
+  port:
+    targetPort: oauth
+  tls:
+    termination: Reencrypt
--- a/config/internal/apiserver/sa_ds-pipeline.yaml.tmpl
+++ b/config/internal/apiserver/sa_ds-pipeline.yaml.tmpl
@@ -3,6 +3,8 @@ kind: ServiceAccount
 metadata:
   name: ds-pipeline-{{.Name}}
   namespace: {{.Namespace}}
+  annotations:
+    serviceaccounts.openshift.io/oauth-redirectreference.primary: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"ds-pipeline-{{.Name}}"}}'
   labels:
     app: ds-pipeline-{{.Name}}
     component: data-science-pipelines
--- a/config/internal/apiserver/service.yaml.tmpl
+++ b/config/internal/apiserver/service.yaml.tmpl
@@ -3,11 +3,17 @@ kind: Service
 metadata:
   name: {{.APIServerServiceName}}
   namespace: {{.Namespace}}
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: ds-pipelines-proxy-tls-{{.Name}}
   labels:
     app: ds-pipeline-{{.Name}}
     component: data-science-pipelines
 spec:
   ports:
+    - name: oauth
+      port: 8443
+      protocol: TCP
+      targetPort: oauth
     - name: http
       port: 8888
       protocol: TCP
--- a/controllers/apiserver.go
+++ b/controllers/apiserver.go
@@ -16,7 +16,10 @@ limitations under the License.
 package controllers
 
 import (
+	"context"
 	dspav1alpha1 "github.com/opendatahub-io/data-science-pipelines-operator/api/v1alpha1"
+	v1 "github.com/openshift/api/route/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 var apiServerTemplates = []string{
@@ -31,7 +34,11 @@ var apiServerTemplates = []string{
 	"apiserver/deployment.yaml.tmpl",
 }
 
-func (r *DSPAReconciler) ReconcileAPIServer(dsp *dspav1alpha1.DataSciencePipelinesApplication, params *DSPAParams) error {
+// serverRoute is a resource deployed conditionally
+// as such it is handled separately
+const serverRoute = "apiserver/route.yaml.tmpl"
+
+func (r *DSPAReconciler) ReconcileAPIServer(ctx context.Context, dsp *dspav1alpha1.DataSciencePipelinesApplication, params *DSPAParams) error {
 
 	if !dsp.Spec.APIServer.Deploy {
 		r.Log.Info("Skipping Application of APIServer Resources")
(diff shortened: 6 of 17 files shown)


config/internal/apiserver/deployment.yaml.tmpl
```diff
@@ -18,6 +18,54 @@ spec:
...
+          args:
+            - --https-address=:8443
+            - --provider=openshift
+            - --openshift-service-account=ds-pipeline-{{.Name}}
+            - --upstream=http://localhost:8888
+            - --tls-cert=/etc/tls/private/tls.crt
+            - --tls-key=/etc/tls/private/tls.key
+            - --cookie-secret=SECRET
+            - '--openshift-delegate-urls={"/": {"group":"route.openshift.io","resource":"routes","verb":"get","name":"ds-pipeline-{{.Name}}","namespace":"{{.Namespace}}"}}'
+            - '--openshift-sar={"namespace":"{{.Namespace}}", "resource": "routes", "resourceName":"ds-pipeline-{{.Name}}", "verb": "get"}'
```
harshad16, line 32: The openshift-sar is used for user authentication,
without the resource group specified, it would think of default resource group "", routes is no longer in default resource group, this would cause any non cluster admin, to have difficulty reaching the service.
```suggestion
            - '--openshift-sar={"namespace":"{{.Namespace}}", "resource": "routes", "resourceName":"ds-pipeline-{{.Name}}", "verb": "get", "resourceAPIGroup":"route.openshift.io"}'
```

or we remove completely the openshift-sar and rely on openshift-delegate, as API server would any way be reached via token most of the times.

harshad16 (commented):
great work ,
just one suggestion

HumairAK:
/hold 

from @rimolive: we should also test using pipeline sdk and confirm nertwork policy doesn't interfere with pipeline creation via api server when we try to use the sdk against oauth proxy.

HumairAK:
/unhold 

Please see the following gif of the sdk being used against the oauth proxy: 


pr-34

rimolive:
> /unhold
> 
> Please see the following gif of the sdk being used against the oauth proxy + networkpolicy in this pr:
> 
> pr-34 [ pr-34 ](https://user-images.githubusercontent.com/10904967/223171187-b1ab625e-1a2e-4258-8c4e-160e92c3b95a.gif) [ ](https://user-images.githubusercontent.com/10904967/223171187-b1ab625e-1a2e-4258-8c4e-160e92c3b95a.gif)
> 
> In this example `testuser` is given this role:
> 
> ```yaml
> kind: Role
> apiVersion: rbac.authorization.k8s.io/v1
> metadata:
>   name: example
> rules:
>   - verbs:
>       - get
>     apiGroups:
>       - route.openshift.io
>     resources:
>       - routes
> ```

Ok, that's enough for DSP users.

config/internal/apiserver/deployment.yaml.tmpl
```diff
@@ -18,6 +18,54 @@ spec:
...
+        {{ if .APIServer.EnableRoute }}
+        - name: oauth-proxy
+          args:
+            - --https-address=:8443
+            - --provider=openshift
+            - --openshift-service-account=ds-pipeline-{{.Name}}
+            - --upstream=http://localhost:8888
+            - --tls-cert=/etc/tls/private/tls.crt
+            - --tls-key=/etc/tls/private/tls.key
+            - --cookie-secret=SECRET
```
gregsheremeta, line 30: Is `--cookie-secret=SECRET` right? Almost looks like a bad copy paste from an example. But it's used in [this example](https://github.com/openshift/oauth-proxy/blob/master/contrib/sidecar.yaml#L65) too. A [different page](https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/overview#generating-a-cookie-secret) suggests it should be "the seed string for secure cookies (optionally base64 encoded)"

config/internal/apiserver/deployment.yaml.tmpl
```diff
@@ -18,6 +18,54 @@ spec:
...
+            - '--openshift-delegate-urls={"/": {"group":"route.openshift.io","resource":"routes","verb":"get","name":"ds-pipeline-{{.Name}}","namespace":"{{.Namespace}}"}}'
```
gregsheremeta, line 31: couldn't quite tell from the docs -- `"/"` means "delegate authn/authz for everything, pass if the user has get access to the route in the namespace ... ?

config/internal/apiserver/deployment.yaml.tmpl
```diff
@@ -18,6 +18,54 @@ spec:
...
+            - '--openshift-sar={"namespace":"{{.Namespace}}","resource":"routes","resourceName":"ds-pipeline-{{.Name}}","verb":"get","resourceAPIGroup":"route.openshift.io"}'
```
gregsheremeta, line 32: I'm not understanding why we need both `delegate-urls` and `sar`. From my brief reading of [this doc](https://github.com/openshift/oauth-proxy#limiting-access-to-users), it sounds like they are competing mechanisms. Maybe I have that wrong.

controllers/apiserver.go
```diff
@@ -31,7 +34,10 @@ var apiServerTemplates = []string{
 	"apiserver/deployment.yaml.tmpl",
 }
 
-func (r *DSPAReconciler) ReconcileAPIServer(dsp *dspav1alpha1.DataSciencePipelinesApplication, params *DSPAParams) error {
+// serverRoute is a Conditional resource
```
gregsheremeta, line 37: what does `Conditional` (capital-C) resource mean?

gregsheremeta (commented):
I didn't finish reviewing specifically the NetworkPolicy, but I figured I'd send these comments off now ... and then I'll do another pass with your answers in hand.

config/internal/apiserver/deployment.yaml.tmpl
```diff
@@ -18,6 +18,54 @@ spec:
...
+              port: oauth
+              scheme: HTTPS
+            initialDelaySeconds: 5
+            timeoutSeconds: 1
+            periodSeconds: 5
+            successThreshold: 1
+            failureThreshold: 3
+          resources:
+            limits:
+              cpu: 100m
```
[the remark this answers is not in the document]
HumairAK: this is what's used by odh-dashboard: 
https://github.com/opendatahub-io/odh-manifests/blob/master/odh-dashboard/base/deployment.yaml#L98-L104 

and previously used by ds-pipelines-ui: 
https://github.com/opendatahub-io/data-science-pipelines/blob/master/manifests/opendatahub/overlays/ds-pipeline-ui/deployments/ds-pipeline-ui.yaml#L58-L63 

and now currently dspo's deployment of ui: 

https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/config/internal/mlpipelines-ui/deployment.yaml.tmpl#L58-L64

config/internal/apiserver/deployment.yaml.tmpl
```diff
@@ -143,3 +191,7 @@ spec:
               {{ end }}
             {{ end }}
       serviceAccountName: ds-pipeline-{{.Name}}
+      volumes:
+        - name: proxy-tls
+          secret:
+            secretName: ds-pipelines-proxy-tls-{{.Name}}
```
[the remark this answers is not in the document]
HumairAK: comes from here: https://github.com/HumairAK/data-science-pipelines-operator/blob/oauth-proxy/config/internal/apiserver/service.yaml.tmpl#L7

ref: 
https://docs.openshift.com/container-platform/4.9/security/certificates/service-serving-certificate.html#add-service-certificate_service-serving-certificate

config/internal/apiserver/route.yaml.tmpl
```diff
@@ -0,0 +1,19 @@
...
+    component: data-science-pipelines
+  annotations:
+    kubernetes.io/tls-acme: "true"
+spec:
+  to:
+    kind: Service
+    name: ds-pipeline-{{.Name}}
+    weight: 100
+  port:
+    targetPort: oauth
```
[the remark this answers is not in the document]
HumairAK: [references the port identified by the service](https://github.com/HumairAK/data-science-pipelines-operator/blob/oauth-proxy/config/internal/apiserver/deployment.yaml.tmpl#L37)

> The target port on pods selected by the service this route points to. If this is a string, it will be looked up as a named port in the target endpoints port list. [[1]](https://docs.openshift.com/container-platform/4.12/rest_api/network_apis/route-route-openshift-io-v1.html)

this is a bit unclear but it's referring to port's named in the service spec

config/internal/apiserver/sa_ds-pipeline.yaml.tmpl
```diff
@@ -3,6 +3,8 @@ kind: ServiceAccount
 metadata:
   name: ds-pipeline-{{.Name}}
   namespace: {{.Namespace}}
+  annotations:
+    serviceaccounts.openshift.io/oauth-redirectreference.primary: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"ds-pipeline-{{.Name}}"}}'
```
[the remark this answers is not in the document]
HumairAK: This is how you can use an SA as an oauth client in ocp: https://docs.openshift.com/container-platform/4.9/authentication/using-service-accounts-as-oauth-client.html#redirect-uris-for-service-accounts_using-service-accounts-as-oauth-client

config/internal/apiserver/route.yaml.tmpl
```diff
@@ -0,0 +1,19 @@
...
+  tls:
+    termination: Reencrypt
```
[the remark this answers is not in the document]
HumairAK: Re-encrypt delegates tls handling in backend to ocp's ingress router, tls-acme is a cert-manager annotation that, in a cluster using cert-manager, can be secured via tls automatically

with re-encrypt + plus passing in cluster certs to the oauth proxy, this prevents issues with deploying this in clusters that are using self-signed certs

config/internal/apiserver/service.yaml.tmpl
```diff
@@ -3,11 +3,17 @@ kind: Service
...
+    service.alpha.openshift.io/serving-cert-secret-name: ds-pipelines-proxy-tls-{{.Name}}
   labels:
     app: ds-pipeline-{{.Name}}
     component: data-science-pipelines
 spec:
   ports:
+    - name: oauth
+      port: 8443
+      protocol: TCP
+      targetPort: oauth
```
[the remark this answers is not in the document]
HumairAK: here it's targetting the pod's port name

config/internal/apiserver/deployment.yaml.tmpl
```diff
@@ -18,6 +18,54 @@ spec:
         component: data-science-pipelines
     spec:
       containers:
+        {{ if .APIServer.EnableRoute }}
+        - name: oauth-proxy
+          args:
```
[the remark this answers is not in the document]
HumairAK: afaik this is the best doc I've been able to come across as well, explaining all args is out of scope for this thread but we can go over these together in slack if you prefer

config/internal/apiserver/deployment.yaml.tmpl
```diff
@@ -18,6 +18,54 @@ spec:
...
+            - --https-address=:8443
+            - --provider=openshift
```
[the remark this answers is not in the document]
HumairAK: specify the openshift provider in oauth proxy: https://github.com/openshift/oauth-proxy/tree/master/providers/openshift

config/internal/apiserver/deployment.yaml.tmpl
HumairAK: `/` refers to the path to which this rule applies to, you can for example restrict it to only paths under `/api`, the rule allows only people that have `get` on `routes` in the ns where this proxy resides, this is no different than how we handled the ui oauth proxy, we expect users with a ds-project to have this permission

config/internal/apiserver/deployment.yaml.tmpl
HumairAK: openshift-delegate-urls >> for  token  based authentication 
openshift-sar >> console/ui based authentication 

@harshad16 can expand more when he's back, as he's recently done deep looking into oauth proxy

controllers/apiserver.go
HumairAK: it just means it's a resource that is only deployed conditionally, and not always, hence why it's not in the array right above it, which are always deployed, I've updated the comments to be more clear here

config/internal/apiserver/deployment.yaml.tmpl
gregsheremeta: you can do a followup for this

controllers/apiserver.go
gregsheremeta: thanks

api/v1alpha1/dspipeline_types.go
```diff
@@ -71,6 +71,9 @@ type APIServer struct {
 	CollectMetrics bool `json:"collectMetrics"`
 	// +kubebuilder:default:=true
 	// +optional
+	EnableRoute bool `json:"enableOauth"`
```
gregsheremeta, line 74: nit: EnableRoute or enableOauth? Feels like a missed replace at some point

PR #35, opened 2023-03-06
use `resource.Quantity` for PVC sizes, CPU, and memory fields

gregsheremeta:
## Description
switch from using `string` to apimachinery `resource.Quantity` for PVC	sizes, CPU, and	memory fields. It's more correct, it gives us free CRD validation, and manifestival seems to be ok with understanding it and applying it.

## How Has This Been Tested?
deployed to a cluster. Created a dspa with defaults. Tried changing `10Gi` to `abc` -- validation error. Tried changing `10Gi` to `11Gi` -- works. Did that with a few different values. Tests all still pass too.

## Merge criteria:


controllers/config/defaults.go
```diff
@@ -59,16 +62,16 @@ const (
...
 
 var (
-	APIServerResourceRequirements         = createResourceRequirement("250m", "500Mi", "500m", "1Gi")
-	PersistenceAgentResourceRequirements  = createResourceRequirement("120m", "500Mi", "250m", "1Gi")
-	ScheduledWorkflowResourceRequirements = createResourceRequirement("120m", "100Mi", "250m", "250Mi")
-	ViewerCRDResourceRequirements         = createResourceRequirement("120m", "100Mi", "250m", "500Mi")
-	MariaDBResourceRequirements           = createResourceRequirement("300m", "800Mi", "1", "1Gi")
-	MinioResourceRequirements             = createResourceRequirement("200m", "100Mi", "250m", "1Gi")
-	MlPipelineUIResourceRequirements      = createResourceRequirement("100m", "256Mi", "100m", "256Mi")
+	APIServerResourceRequirements         = createResourceRequirement(resource.MustParse("250m"), resource.MustParse("500Mi"), resource.MustParse("500m"), resource.MustParse("1Gi"))
```
gregsheremeta, line 65: I'm not sure if this is idiomatic go or not. Feel free to suggest a better way.

controllers/dspipeline_params.go
```diff
@@ -245,7 +247,7 @@ func (p *DSPAParams) SetupObjectParams(ctx context.Context, dsp *dspa.DataScienc
 		// populated with defaults.
 		setStringDefault(config.MinioImage, &p.Minio.Image)
 		setStringDefault(config.MinioDefaultBucket, &p.Minio.Bucket)
-		setStringDefault(config.MinioPVCSize, &p.Minio.PVCSize)
+		p.Minio.PVCSize = resource.MustParse(config.MinioPVCSize)
```
HumairAK, line 250: 2 things here: 

1. we want to only set default if user has not specified a PVC Size in the CR, atm we always use default, see `setStringDefault`, we may want something similar for resource struct 

2. We also want to do something similar for mariaDB. At the moment it will still work without something like this for mariadb because [we have default via kubebuilder annotation](https://github.com/opendatahub-io/data-science-pipelines-operator/pull/35/files#diff-4606cde41444090b0c00d3e515fee3ac2e1ce6de20de579165193a861175a6acR138). If we want to rely on this default, then we should be consistent and take this out of the minio logic here. If we want to ensure we check for defaults in code, and not just rely on k8s to ensure defaults on it send, then add default check in mariadb.

controllers/config/defaults.go
HumairAK: change set can be shorter by just changing the struct builder: 

```golang
func createResourceRequirement(RequestsCPU string, RequestsMemory string, LimitsCPU string, LimitsMemory string) dspav1alpha1.ResourceRequirements {
	return dspav1alpha1.ResourceRequirements{
		Requests: &dspav1alpha1.Resources{
			CPU:    resource.MustParse(RequestsCPU),
			Memory: resource.MustParse(RequestsMemory),
		},
		Limits: &dspav1alpha1.Resources{
			CPU:    resource.MustParse(LimitsCPU),
			Memory: resource.MustParse(LimitsMemory),
		},
	}
}
```

gregsheremeta: From my reading of the [doc](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource) and the examples provided, it seems unsafe to put the `MustParse` call into the function. Doc says:
"MustParse turns the given string into a quantity or panics; for tests or other cases where you know the string is valid." Note that it doesn't return an err -- it just panics.

I admit that it's unlikely that we'll ever call the function with anything other than a valid Quantity string, but forcing callers to call MustParse themselves felt a little safer to me.

If you still think it's better to change it, I'll change it.

controllers/dspipeline_params.go
gregsheremeta: changed it to rely on the kubebuilder annotation

HumairAK:
e2e keeps failing on different points in time that are not related to the tests, and seem to be infra related. The last iteration passed the tests but failed at must-gather so I'm approving the pr and overriding the e2e, logs on e2e passing in the last iteration: 
```
INFO[2023-03-20T23:31:14Z] Step dsp-operator-e2e-dsp-operator-e2e succeeded after 8m38s.
INFO[2023-03-20T23:31:14Z] Step phase test succeeded after 8m38s.
```

And failure on must-gather (happens after tests are done) 

```
Error running must-gather collection: creating temp namespace: Post "https://api.ci-op-wsr7jk0l-9de77.origin-ci-int-aws.dev.rhcloud.com:6443/api/v1/namespaces": dial tcp: lookup [api.ci-op-wsr7jk0l-9de77.origin-ci-int-aws.dev.rhcloud.com](http://api.ci-op-wsr7jk0l-9de77.origin-ci-int-aws.dev.rhcloud.com/) on 172.30.0.10:53: no such host
```

/override  ci/prow/dsp-operator-e2e 
/approve

PR #46, opened 2023-03-10
WIP: add test case that ensures CR options have priority

gregsheremeta:
WIP

I noticed that the DSPA CR APIServer.Image field is always ignored. This test proves it out.

## Merge criteria:

### DIFF
--- a/controllers/testdata/results/case_4/apiserver/deployment.yaml
+++ b/controllers/testdata/results/case_4/apiserver/deployment.yaml
@@ -1,0 +1,187 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ds-pipeline-testdsp4
+  namespace: default
+  labels:
+    app: ds-pipeline-testdsp4
+    component: data-science-pipelines
+spec:
+  selector:
+    matchLabels:
+      app: ds-pipeline-testdsp4
+      component: data-science-pipelines
+  template:
+    metadata:
+      labels:
+        app: ds-pipeline-testdsp4
+        component: data-science-pipelines
+    spec:
+      containers:
+        - name: oauth-proxy
+          args:
+            - --https-address=:8443
+            - --provider=openshift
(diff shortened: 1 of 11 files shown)


controllers/testdata/results/case_4/apiserver/deployment.yaml
```diff
@@ -0,0 +1,187 @@
...
+              value: "false"
+            - name: MINIO_SERVICE_SERVICE_HOST
+              value: "minio-testdsp4.default.svc.cluster.local"
+            - name: MINIO_SERVICE_SERVICE_PORT
+              value: "9000"
+            - name: CACHE_IMAGE
+              value: "ubi-minimal:test4"
+            - name: MOVERESULTS_IMAGE
+              value: "busybox:test4"
+          image: this-apiserver-image-from-cr-should-be-used:test4
```
gregsheremeta, line 141: @HumairAK this will fail, but if I change the image here to `this-apiserver-image-from-config-should-not-be-used:test4` it will pass.

I don't see the field being read out of the DSPA CR [here](https://github.com/opendatahub-io/data-science-pipelines-operator/blob/b45d61f95586bc007f898da6010d0757916b9e26/controllers/dspipeline_params.go#L359).

Is this a bug, or am I missing something?

gregsheremeta: In other words, I thought that the config was a configmap for defaults that should only be used if something wasn't specified in the CR. It was our way to un-hardcode defaults. But it looks to me like those defaults from the configmap (for this apiserver image, at least) are always used no matter what's in the CR.

HumairAK: definitely a bug, CR should take precedence

gregsheremeta: thanks for confirming!

gregsheremeta:
closing until someone works on #67

PR #47, opened 2023-03-10
add kubebuilder validation markers for DSPA

gregsheremeta:
## Description
add kubebuilder validation markers for DSPA. There's currently only one field that's validate-able via simple markers.

## How Has This Been Tested?
Deploy to a cluster and try to set something not in the enum, and it works as expected

## Merge criteria:

### DIFF
--- a/config/crd/bases/datasciencepipelinesapplications.opendatahub.io_datasciencepipelinesapplications.yaml
+++ b/config/crd/bases/datasciencepipelinesapplications.opendatahub.io_datasciencepipelinesapplications.yaml
@@ -105,6 +105,10 @@ spec:
                     type: boolean
                   terminateStatus:
                     default: Cancelled
+                    enum:
+                    - Cancelled
+                    - StoppedRunFinally
+                    - CancelledRunFinally
                     type: string
                   trackArtifacts:
                     default: true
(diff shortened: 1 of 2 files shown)


config/crd/bases/datasciencepipelinesapplications.opendatahub.io_datasciencepipelinesapplications.yaml
```diff
@@ -105,6 +105,10 @@ spec:
                     type: boolean
                   terminateStatus:
                     default: Cancelled
+                    enum:
+                    - Cancelled
```
gregsheremeta, line 109: these are from [this kfp-tekton doc](https://github.com/kubeflow/kfp-tekton/blob/eb44737195097d1880a63dbc299f4dcc89e27e6e/guides/kfp-admin-guide.md#change-the-tekton-terminate-api-method)

kinda hate to hardcode them, but the only other thing i can think of is to not use the markers :(

HumairAK: crd is auto generated, this will end up getting replaced

gregsheremeta: sorry, I can see how my comment was confusing. I didn't touch this file. Its contents are from `make generate`

I should have commented in the types file. What I meant was I didn't care for hardcoding the three values in an annotation. I'd prefer to include them somehow from kfp-tekton code, but I don't see an easy way to do that. In other words, when kfp-tekton adds a new valid option for this, we're broken.

PR #73, opened 2023-03-24
Remove redundant/unused code.

HumairAK:
## Description
Some cleanup, removes redundant if statements and unused function parameters.

## How Has This Been Tested?
Unit/Integration tests should be sufficient for this, as there's no change to logic.

## Merge criteria:

### DIFF
--- a/controllers/dspipeline_controller.go
+++ b/controllers/dspipeline_controller.go
@@ -62,10 +62,7 @@ func (r *DSPAReconciler) Apply(owner mf.Owner, params *DSPAParams, template stri
 		return err
 	}
 
-	if err = tmplManifest.Apply(); err != nil {
-		return err
-	}
-	return nil
+	return tmplManifest.Apply()
 }
 
 func (r *DSPAReconciler) ApplyWithoutOwner(params *DSPAParams, template string, fns ...mf.Transformer) error {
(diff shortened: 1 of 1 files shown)


controllers/dspipeline_controller.go
```diff
@@ -62,10 +62,8 @@ func (r *DSPAReconciler) Apply(owner mf.Owner, params *DSPAParams, template stri
 		return err
 	}
 
-	if err = tmplManifest.Apply(); err != nil {
-		return err
-	}
-	return nil
+	err = tmplManifest.Apply()
```
gmfrasca, line 65: could you simplify this even further with just `return tmplManifest.Apply()`?  Or is that breaking conventions?

controllers/dspipeline_controller.go
```diff
@@ -259,7 +253,7 @@ func (r *DSPAReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 // Clean Up any resources not handled by garbage collection, like Cluster ResourceRequirements
-func (r *DSPAReconciler) cleanUpResources(ctx context.Context, req ctrl.Request, dsp *dspav1alpha1.DataSciencePipelinesApplication, params *DSPAParams) error {
+func (r *DSPAReconciler) cleanUpResources(params *DSPAParams) error {
 	err := r.CleanUpCommon(params)
```
gmfrasca, line 257: since we're simplifying here, could we also just return the result here instead of checking if `err` is nil, just to return `nil` if it is?

gmfrasca (commented):
/lgtm

DharmitD (approved):
/Approve

anishasthana (approved):

PR #99, opened 2023-04-24
Fix: only enable tag-release-quay workflow for upstream opendatahub-io

zdtsw:
- currently any forked repo will be triggered by new tag and failure

When I created new [tag](https://github.com/red-hat-data-services/data-science-pipelines-operator/releases/tag/v1.25.0-rhods) in downstream RHODS repo, I got failed run https://github.com/red-hat-data-services/data-science-pipelines-operator/actions/runs/4783277428
- the reason is it does not have quay auth
- but we should not need this run at all

## Description
This PR is only adding an extra check to see if the repo is from opendatahub-io or not 
It prevents any forked repo to start workflow by any new tag

## How Has This Been Tested?

- Manually created a tag named "test-tag" locally
- Push to my forked github repo, see https://github.com/zdtsw/data-science-pipelines-operator/tags
- No Action was started, see https://github.com/zdtsw/data-science-pipelines-operator/actions

## Merge criteria:

### DIFF
--- a/.github/workflows/tag-release-quay.yml
+++ b/.github/workflows/tag-release-quay.yml
@@ -1,6 +1,6 @@
 name: Image push per Github Tag
 
-# This GitHub action activates whenever a new tag is created on the repo
+# This GitHub action activates whenever a new tag is created on the repo under "opendatahub-io"
 # and creates a copy of the image of the associated commit hash with the
 # appropriate tag name.
 
@@ -14,6 +14,7 @@ env:
 jobs:
   copy-tag-to-quay:
     runs-on: ubuntu-latest
+    if: github.repository == 'opendatahub-io/data-science-pipelines-operator'
     steps:
       - name: Git checkout
         uses: actions/checkout@v3


.github/workflows/tag-release-quay.yml
```diff
@@ -14,6 +14,7 @@ env:
 jobs:
   copy-tag-to-quay:
     runs-on: ubuntu-latest
+    if: 'startsWith(${{ github.event.repository.name }}, "opendatahub-io")'
```
harshad16, line 17:
```suggestion
    if: github.repository == 'opendatahub-io/data-science-pipelines-operator'
```

harshad16 (changes requested):
The expression doesn't work for me.


Tried,
including the change 
https://github.com/harshad16/data-science-pipelines-operator/blob/82ec6a0c99edcb2128a7969c329110d2fef925e9/.github/workflows/tag-release-quay.yml#L17
and action triggers:
https://github.com/harshad16/data-science-pipelines-operator/actions/runs/4787736035

.github/workflows/tag-release-quay.yml
zdtsw: sure, if this makes more clear with intention !

PR #108, opened 2023-04-26
Make object store port/secure scheme optional.

HumairAK:
## Description

For cloud storage like aws s3, the secure port is 443, but this does not work with the minio module that is used by dsp api server, so instead in such a case we make port optional, and do not require it. When an empty port is provided, then it is simply not used by dsp api server.

Test cases are provided to reflect this change.

Docs have also been provided on how to utilize this external object storage with an aws s3 specific example usecase.

## How Has This Been Tested?
Locally and in cluster. 

## Merge criteria:


gmfrasca (commented):
/lgtm

PR #110, opened 2023-05-01
Update DSP Operator labels

DharmitD:
## Description

Update DSP Operator labels to be more specific to DSP/DSPO. Closes [RHODS-7918](https://issues.redhat.com/browse/RHODS-7918)

## How Has This Been Tested?

Deployed a DSPO instance, checked that the deployment, service, service monitor, roles came up correctly.
Checked that the metrics are listed out in the `Observe` section of the OpenShift console:
Screenshot from 2023-05-08 11-38-34

## Merge criteria:

### DIFF
--- a/config/manager/manager-service.yaml
+++ b/config/manager/manager-service.yaml
@@ -3,11 +3,10 @@ kind: Service
 metadata:
   name: service
   labels:
-    control-plane: controller-manager
-    app.kubernetes.io/created-by: data-science-pipelines-operator
+    app.kubernetes.io/name: data-science-pipelines-operator
 spec:
   ports:
     - name: metrics
       port: 8080
   selector:
-    control-plane: controller-manager
+    app.kubernetes.io/name: data-science-pipelines-operator
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -4,24 +4,18 @@ metadata:
   name: controller-manager
   namespace: datasciencepipelinesapplications-controller
   labels:
-    control-plane: controller-manager
-    app.kubernetes.io/name: deployment
-    app.kubernetes.io/instance: controller-manager
-    app.kubernetes.io/component: manager
-    app.kubernetes.io/created-by: data-science-pipelines-operator
-    app.kubernetes.io/part-of: data-science-pipelines-operator
+    app.kubernetes.io/name: data-science-pipelines-operator
 spec:
   selector:
     matchLabels:
-      control-plane: controller-manager
+      app.kubernetes.io/name: data-science-pipelines-operator
   replicas: 1
   template:
     metadata:
       annotations:
         kubectl.kubernetes.io/default-container: manager
       labels:
-        control-plane: controller-manager
-        app.kubernetes.io/created-by: data-science-pipelines-operator
+        app.kubernetes.io/name: data-science-pipelines-operator
     spec:
       securityContext:
         runAsNonRoot: true
--- a/config/prometheus/monitor.yaml
+++ b/config/prometheus/monitor.yaml
@@ -9,4 +9,4 @@ spec:
       port: metrics
   selector:
     matchLabels:
-      control-plane: controller-manager
+      app.kubernetes.io/name: data-science-pipelines-operator
--- a/config/rbac/leader_election_role.yaml
+++ b/config/rbac/leader_election_role.yaml
@@ -3,11 +3,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   labels:
-    app.kubernetes.io/name: role
-    app.kubernetes.io/instance: leader-election-role
-    app.kubernetes.io/component: rbac
-    app.kubernetes.io/created-by: data-science-pipelines-operator
-    app.kubernetes.io/part-of: data-science-pipelines-operator
+    app.kubernetes.io/name: data-science-pipelines-operator
   name: leader-election-role
 rules:
 - apiGroups:
--- a/config/rbac/leader_election_role_binding.yaml
+++ b/config/rbac/leader_election_role_binding.yaml
@@ -2,11 +2,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   labels:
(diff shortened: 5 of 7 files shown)


config/manager/manager-service.yaml
```diff
@@ -3,11 +3,15 @@ kind: Service
 metadata:
   name: service
   labels:
-    control-plane: controller-manager
-    app.kubernetes.io/created-by: data-science-pipelines-operator
+    app.kubernetes.io/name: service
```
VannTen, line 6: I think this is too granular:
The [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) in Kubernetes are supposed to be about application, not individual resources (otherwise they'd be redondant).
Here we would rather have either "controller-manager" (if we consider the controller as an application itself, part of DSPO), or just "DSPO" / "data-science-pipelines-operator".

config/manager/manager-service.yaml
```diff
@@ -3,11 +3,15 @@ kind: Service
...
+    app.kubernetes.io/component: manager
+    app.kubernetes.io/part-of: data-science-pipelines-operator
+    app.kubernetes.io/instance: controller-manager
```
VannTen, line 9: This should have some part of unique id per installation, so that should probably be left to the deployer.

config/manager/manager-service.yaml
```diff
@@ -3,11 +3,15 @@ kind: Service
 metadata:
   name: service
   labels:
-    control-plane: controller-manager
```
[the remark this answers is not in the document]
DharmitD: Tested this out in metrics stack, ServiceMonitor works as expected with the new label[ here](https://github.com/opendatahub-io/data-science-pipelines-operator/pull/110/commits/1256cbae8ecd3449aeaf1d3f0dce23342318c7e4#diff-06cc310278af3cef0bcd13ae45e20a4f900447dca3f7451a0719a7522bb2b39bR12)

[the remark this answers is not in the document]
DharmitD: There is just one more Service Monitor in this repository, which is for the DSP Application. That one has[ a different matchlabel ](https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/config/internal/apiserver/monitor.yaml.tmpl#L15) anyway.

config/manager/manager-service.yaml
DharmitD: By unique id do you mean a version id? What could the version id be based on? 

If this would be something based on versions of DSP, then I think we would need to update it for every version. Is this label really required then?

config/manager/manager-service.yaml
DharmitD: Makes sense, would you suggest the name label be just "data-science-pipelines-operator" ?

DharmitD:
> The examples in https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ are quite insightful.
> 
> I think we should loosely be able to say something '<app.kubernetes.io/name> is the <app.kubernetes.io/component> in the <app.kubernetes.io/part-of> application'
> 
> (In the K8s doc example that would give : 'mysql is the database in wordpress application'.
> 
> Wdyt ?

The example refers to a specific _application_, which is why the analogy of '<app.kubernetes.io/name> is the <app.kubernetes.io/component> in the <app.kubernetes.io/part-of> application' is applicable. I'm not sure how this analogy would fit in with different resources in an _operator_.

If we were to follow it, then in our (Data Science Pipelines Operator's) case, there could just be one label - `name`. Something like: `app.kubernetes.io/name: data-science-pipelines-operator`, across all the resources. Does that make sense?

I think the whole idea behind revamping labels is to have labels that correctly filter DSPO resources. Even if it's a single, we should be good. What do you think?

cc: @HumairAK @rimolive @gmfrasca @gregsheremeta

config/manager/manager-service.yaml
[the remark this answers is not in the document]
VannTen: Great, disregard this part then. I've seen too much ServiceMonitor recently, that was probably somewhere else.

config/manager/manager-service.yaml
VannTen: Not version id, because if we follow the conventions of that label, it should be distinct between two different installation (even of the same version). That's why I think we should scrap it out of the yaml, only the installer has the capacity to generate and apply a random id for each instance (maybe that's doable with kustomize generator stuff, but I'm not too sure)

config/manager/manager-service.yaml
VannTen: I think so. In that case I we consider the application is "data-science-pipelines-operator" which is itself "part-of" ODH.

The other alternative: the overall product ("part-of") is DSPO, but in that case I'm not sure what should be in "name".
-> which probably hints that the first alternative is better.

VannTen:
The single label sounds fine, but maybe the part-of could be ODH/OpendataHub (whatever spelling fits) ?

(we can also scrap the part-of, and maybe we'll have an upper level kustomization apply it).

HumairAK:
> I think the whole idea behind revamping labels is to have labels that correctly filter DSPO resources. Even if it's a single label, we should be good. What do you think?

I'm fine with this, afaik the original Jira scop is to correctly identify DSPO and its resources uniquely, `controller-manager` is very vague, and it's the default added to an operator via operator-sdk, so selecting a service on this label will likely result in issues. Broader discussions on label architecture are a bit out of scope for this. Let's get minimum labels needed to uniquely identify DSPO and select on its service.

DharmitD:
> > I think the whole idea behind revamping labels is to have labels that correctly filter DSPO resources. Even if it's a single label, we should be good. What do you think?
> 
> I'm fine with this, afaik the original Jira scope is to correctly identify DSPO and its resources uniquely, `controller-manager` is very vague, and it's the default added to an operator via operator-sdk, so selecting a service on this label will likely result in issues. Broader discussions on label architecture are a bit out of scope for this. Let's get minimum labels needed to uniquely identify DSPO and select on its service.

Updated to have a single label across all DSPO resources as per our discussion here.

VannTen:
/lgtm
Way cleaner that way :+1:

PR #120, opened 2023-05-17
Parameterize the oauth-proxy image as config var

harshad16:
Parameterize the oauth-proxy image as config var

## Description

Move the oauth-proxy image to configmap , so it can assigned value directly.

## How Has This Been Tested?

Use the below image for the deployment
Image: `quay.io/opendatahub/data-science-pipelines-operator:pr-120`
replace the value here:https://github.com/opendatahub-io/data-science-pipelines-operator/blob/fe12ad963f0c7f498c941f364ffccbe0985b53c0/config/base/params.env#L8

deploy the operator and check the yaml for testing.

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #126, opened 2023-05-23
Bugfix: Images Specified in DSPA CRs should override DSPO defaults

### LINKED ISSUE #67: CR Images are not being applied due to config overriding
ATM None of the image fields in the CR work because the images provided via the config are _always_ preferred. We need to fix this such that when an image for _any_ component with a configurable image field (so not including UI and minio) is set by the CR, it is preferred over default configurations. 

Context [here](https://github.com/opendatahub-io/data-science-pipelines-operator/pull/46#discussion_r1132895109).

gmfrasca:
Fixes #67 

If the `image` field is specified in a DSPA CR, it should take precedence over the image defaults provided in the DSPO config.

## Description
Changes DSPAParams behavior to only set the XXXX.Image component if it is not specified in the DSPA CR

## How Has This Been Tested?
- Deploy operator built from PR.
- Create DSPA from stardard dspa simple, verify 'normal' images used
- Create DSPA with `image` field specified for APIServer, PersistenceAgent, ScheduledWorkflow, and UI.  Also set ArtifactImage, CacheImage, and MoveResultImage in apiServer spec.  These can be fake values.  Verify the specified images have overrident the ones used from first DSPA

## Merge criteria:

### DIFF
--- a/controllers/testutil/equalities.go
+++ b/controllers/testutil/equalities.go
@@ -18,11 +18,12 @@ package testutil
 
 import (
 	"fmt"
+	"reflect"
+
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/kubernetes/scheme"
-	"reflect"
 )
 
 // CompareResourceProcs maps object kind's to their associated comparison procedures.
@@ -119,7 +120,7 @@ func deploymentsAreEqual(expected, actual *unstructured.Unstructured) (bool, err
 		for i, expectedEnv := range expectedContainer.Env {
 			actualEnv := actualContainer.Env[i]
 			if !reflect.DeepEqual(expectedEnv, actualEnv) {
-				return false, notEqualMsg(fmt.Sprintf("Container Env [expected: %s, actual: %s]", expectedEnv.Name, actualEnv.Name))
+				return false, notEqualMsg(fmt.Sprintf("Container Env [expected: %s=%s, actual: %s=%s]", expectedEnv.Name, expectedEnv.Value, actualEnv.Name, actualEnv.Value))
 			}
 		}
 
@@ -139,7 +140,7 @@ func deploymentsAreEqual(expected, actual *unstructured.Unstructured) (bool, err
 			return false, notEqualMsg("Container Name")
 		}
 		if expectedContainer.Image != actualContainer.Image {
-			return false, notEqualMsg("Container Image")
+			return false, notEqualMsg(fmt.Sprintf("Container Image [expected: %s, actual: %s]", expectedContainer.Image, actualContainer.Image))
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


HumairAK:
@gmfrasca can you rebase, so you can incorporate https://github.com/opendatahub-io/data-science-pipelines-operator/pull/120

gmfrasca:
For testing, use this DSPA definition:
```yaml
metadata:
  name: badimages 
spec:
  apiServer:
    deploy: true
    image: quay.io/modh/nonexistant-apiserver:v0.0.0
    artifactImage: quay.io/modh/nonexistant-artifactmangager:v0.0.0
    cacheImage: registry.access.redhat.com/ubi8/nonexistant-ubi-minimal:v0.0.0
    moveResultsImage: registry.access.redhat.com/ubi8/nonexistant-ubi-minimal:v1.1.1
  persistenceAgent:
    deploy: true
    image: quay.io/modh/nonexistant-persistenceagent:v0.0.0
  scheduledWorkflow:
    deploy: true
    image: quay.io/modh/nonxistant-scheduledworkflow:v0.0.0
  mlpipelineUI:
    deploy: true
    image: quay.io/opendatahub/nonexistant-dspui:v0.0.0
  database:
    mariaDB:
      deploy: true
      image: registry.redhat.io/rhel8/nonexistant-mariadb:v0.0.0
  objectStorage:
    minio:
      deploy: true
      image: quay.io/opendatahub/nonexistant-minio:v0.0.0
```

controllers/testutil/equalities.go
```diff
@@ -119,7 +120,7 @@ func deploymentsAreEqual(expected, actual *unstructured.Unstructured) (bool, err
 		for i, expectedEnv := range expectedContainer.Env {
 			actualEnv := actualContainer.Env[i]
 			if !reflect.DeepEqual(expectedEnv, actualEnv) {
-				return false, notEqualMsg(fmt.Sprintf("Container Env [expected: %s, actual: %s]", expectedEnv.Name, actualEnv.Name))
+				return false, notEqualMsg(fmt.Sprintf("Container Env [expected: %s=%s, actual: %s=%s]", expectedEnv.Name, expectedEnv.Value, actualEnv.Name, actualEnv.Value))
```
HumairAK, line 123: `EnvVar.Value` is an optional field, sometimes it can be `ValueFrom` so if you want to show the value I think you need to check for the case that it's not `""`

HumairAK: https://pkg.go.dev/k8s.io/api@v0.27.2/core/v1#EnvVar

gmfrasca: if it's default value is "", wouldn't this just print something like
`Container Env [expected: foo=bar, actual: foo=]`

imo that's fine and not worth complicating this block with conditionals just for logging

HumairAK: yeah it would, I think it's a bit misleading if the actual is "ValueFrom", in such a case the expectedEnv.Value may also be different (empty if we're testing valuefrom), but I don't feel strong about this, just pointing it out.

PR #128, opened 2023-05-26
default to deploying dspo's main tag

gregsheremeta:
I noticed when deploying DSPO by pointing a kfdef at this repository, I get a month-old version of DSPO. When I do the same for dashboard, I get main. We might as well default to main too ... if someone is pointing at DSPO or building it from source, they probably want bleeding edge anyway.

## How Has This Been Tested?
I didn't test it yet.

## Merge criteria:


HumairAK:
@gregsheremeta would you be willing to also set the `imagePullPolicy for the container spec to `Always`, otherwise people deploying `main` tag may end up deploying older versions because of cached images in the registry

HumairAK:
>If we hardcode it to Always, it will be Always in odh-manifests too, despite us not using main there. 

That's a good point, let's leave it.
/lgtm
/approve

PR #133, opened 2023-06-05
Add Database and ObjectStorage Health Checks

### LINKED ISSUE #149: Add Logic to DSP Operator to wait for s3/db to be ready before deploying
Task should also update CR status field for DSPO when s3/db are ready/not ready accordingly.

gmfrasca:
Fixes #149

- Delays deploying DSPA Manifests until DB and ObjStore Connections are healthy

## Description
- Performs simple health checks on the database and object storage connections
- Prevents DSPO from deploying DSPA components (apiserver, persistenceagent, etc) until conns are healthy
- Introduces DatabaseAvailable and ObjectStoreAvailable readiness metrics

## How Has This Been Tested?

**Notes to reviewer**:

- ~In it's current state, this makes 'local' runs of the DSPO difficult/impossible, as the Object Store host references internal Services which are not available on an external-running DSPO~
- ~Do we want to add an override mechanism here?~
- Update: a `spec.database.disableHealthCheck` field and funtionality was added to the DSPA CRD to enable local runs

## Merge criteria:


controllers/metrics.go
```diff
@@ -23,6 +23,26 @@ import (
 
 // Prometheus metrics gauges
 var (
+	DBAvailableMetric = prometheus.NewGaugeVec(
```
HumairAK, line 26: Don't you need to [register these like the others](https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/controllers/metrics.go#L69)?

controllers/dspipeline_controller.go
```diff
@@ -397,7 +409,16 @@ func (r *DSPAReconciler) handleReadyCondition(
...
+func (r *DSPAReconciler) GenerateStatus(ctx context.Context, dspa *dspav1alpha1.DataSciencePipelinesApplication, params *DSPAParams) ([]metav1.Condition, error) {
+	databaseAvailable := r.buildCondition(config.DatabaseAvailable, dspa, config.DatabaseAvailable)
+	if r.isDatabaseAccessible(ctx, dspa, params) {
+		databaseAvailable.Status = metav1.ConditionTrue
+	}
+
+	objStoreAvailable := r.buildCondition(config.ObjectStoreAvailable, dspa, config.ObjectStoreAvailable)
+	if r.isObjectStorageAccessible(ctx, dspa, params) {
+		objStoreAvailable.Status = metav1.ConditionTrue
+	}
```
HumairAK, line 421: It looks like we're doing this call twice, can we just pass the values we obtained [here](https://github.com/gmfrasca/data-science-pipelines-operator/blob/6992e9b4678a973e529d5a10c5382645f20089d8/controllers/dspipeline_controller.go#L224)?

controllers/dspipeline_controller.go
```diff
@@ -255,7 +264,7 @@ func (r *DSPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{}, err
 	}
 
-	conditions, err := r.GenerateStatus(ctx, dspa)
+	conditions, err := r.GenerateStatus(ctx, dspa, params)
```
HumairAK, line 267: When deploying the DSPA for the first time, I don't see Statuses show up until the default mariadb/minio are deployed. This is problematic because ODH Dashboard relies on the "Ready" condition type to assess the current status of the DSPA.

controllers/dspipeline_controller.go
```diff
@@ -397,7 +409,16 @@ func (r *DSPAReconciler) handleReadyCondition(
 
 }
 
-func (r *DSPAReconciler) GenerateStatus(ctx context.Context, dspa *dspav1alpha1.DataSciencePipelinesApplication) ([]metav1.Condition, error) {
+func (r *DSPAReconciler) GenerateStatus(ctx context.Context, dspa *dspav1alpha1.DataSciencePipelinesApplication, params *DSPAParams) ([]metav1.Condition, error) {
+	databaseAvailable := r.buildCondition(config.DatabaseAvailable, dspa, config.DatabaseAvailable)
+	if r.isDatabaseAccessible(ctx, dspa, params) {
+		databaseAvailable.Status = metav1.ConditionTrue
```
HumairAK, line 415: Can we also provide meaningful [Messages](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L742) for the condition

controllers/dspipeline_controller.go
```diff
@@ -271,24 +280,20 @@ func (r *DSPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 // handleReadyCondition evaluates if condition with "name" is in condition of type "conditionType".
 // this procedure is valid only for conditions with bool status type, for conditions of non bool type
 // results are undefined.
-func (r *DSPAReconciler) handleReadyCondition(
-	ctx context.Context,
-	dspa *dspav1alpha1.DataSciencePipelinesApplication,
-	name string,
-	condition string,
-) (metav1.Condition, error) {
+func (r *DSPAReconciler) handleReadyCondition(ctx context.Context, dspa *dspav1alpha1.DataSciencePipelinesApplication, name string, condition string) (metav1.Condition, error) {
```
HumairAK, line 296: how come you made this into one line? looks cleaner the other way no?

controllers/dspipeline_controller.go
```diff
@@ -297,7 +302,14 @@ func (r *DSPAReconciler) handleReadyCondition(
 
 	err := r.Get(ctx, types.NamespacedName{Name: component, Namespace: dspa.Namespace}, deployment)
 	if err != nil {
-		return metav1.Condition{}, err
+		if apierrs.IsNotFound(err) {
+			readyCondition.Reason = config.ComponentDeploymentNotFound
+			readyCondition.Status = metav1.ConditionFalse
+			readyCondition.Message = fmt.Sprintf("Deployment for component \"%s\" is missing", component)
```
HumairAK, line 308: The message sounds like something went wrong, but it seems expected behavior to see this when we're waiting for db or storage to come up right? I'm just wondering if we could word this differently to sound less ominous, like "Deployment for component is waiting for pre-requisites to come up,..."

controllers/database.go
```diff
@@ -31,6 +36,53 @@ var dbTemplates = []string{
...
+			log.Info("Database Health Check Successful")
+		} else {
+			log.Info("Unable to connect to Database")
+		}
+		return db_connect
+
+	}
+
+	log.Info("Could not connect to Database: Unsupported Type")
+	// Only MariaDB and Mysql-Compliant Database supported.
```
HumairAK, line 82: this comment looks like something that doesn't belong here, and is better added [here](https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/api/v1alpha1/dspipeline_types.go#L122). 

If you really want to add it here, then I think we should add it as a function  docstring/comment right above it.

controllers/dspipeline_controller.go
gmfrasca: linter yelled at me :)

controllers/database.go
```diff
@@ -31,6 +36,53 @@ var dbTemplates = []string{
...
+	}
+	defer db.Close()
+
+	testStatement := "SELECT 1;"
+	_, err = db.Exec(testStatement)
+	return err == nil
+}
+
+func (r *DSPAReconciler) VerifyMySQLDBConnection(host, port, username, password, dbname string) bool {
+	return ConnectAndQueryDatabase(host, port, username, password, dbname)
```
HumairAK, line 54: can you not just remove the receiver on `VerifyMySQLDBConnection`? doesn't seem like it needs to be a method on DSPAReconciler, so you can mock it directly, unless I missed something.

controllers/storage.go
```diff
@@ -18,7 +18,13 @@ package controllers
 
 import (
 	"context"
+	"encoding/base64"
+	"fmt"
+	"net/http"
 
+	"github.com/go-logr/logr"
+	minio "github.com/minio/minio-go/v7"
```
HumairAK, line 26: `minio` alias is redundant here you can remove it

gmfrasca: DSP/KFP still define the alias this way:
https://github.com/opendatahub-io/data-science-pipelines/blob/master/backend/src/apiserver/storage/object_store.go#L24

controllers/dspipeline_controller.go
gmfrasca: sure, but was trying to avoid it as it pollutes the function signatures.  That said, these are 'expensive' calls so probably better to pass them in in this case

controllers/dspipeline_controller.go
```diff
@@ -420,7 +444,7 @@ func (r *DSPAReconciler) GenerateStatus(ctx context.Context, dspa *dspav1alpha1.
 	crReady := r.buildCondition(config.CrReady, dspa, config.MinimumReplicasAvailable)
 	crReady.Type = config.CrReady
 
-	componentConditions := []metav1.Condition{apiServerReady, persistenceAgentReady, scheduledWorkflowReady}
+	componentConditions := []metav1.Condition{databaseAvailable, objStoreAvailable, apiServerReady, persistenceAgentReady, scheduledWorkflowReady}
```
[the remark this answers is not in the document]
gmfrasca: the problem outlined in `2.` is likely because the db/objstore `Message` field is not set on failure, so the list of failures is just two empty strings concatted.

controllers/metrics.go
gmfrasca: done

controllers/dspipeline_controller.go
gmfrasca: added

controllers/dspipeline_controller.go
gmfrasca: updated

controllers/dspipeline_controller.go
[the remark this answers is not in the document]
gmfrasca: 1. implemented a `disableHealthCheck` override in the CR Spec.

2. Confirmed this was from 2 empty Messages being aggregated together with newline characters only. (the `2` in `|2` is an indentation indicator, and there are 2 newlines following, so the equivalent string being set here was `  \n  \n`.  Adding status messages resolved this.

controllers/dspipeline_controller.go
gmfrasca: updated

controllers/database.go
gmfrasca: moved the conditionals around, so this code and comment now should make a litle more sense

controllers/database.go
gmfrasca: yep, already was doing this with the respective Object Store side. updated

controllers/storage.go
```diff
@@ -31,6 +37,78 @@ var storageTemplates = []string{
...
 }
 
+func joinHostPort(host, port string) string {
+	if port == "" {
+		return host
+	}
+	return fmt.Sprintf("%s:%s", host, port)
+}
+
+func createCredentialProvidersChain(endpoint, accessKey, secretKey string) *credentials.Credentials {
```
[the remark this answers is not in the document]
gmfrasca: removed

controllers/storage.go
gmfrasca: updated =/

controllers/dspipeline_controller.go
```diff
@@ -217,34 +218,42 @@ func (r *DSPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 	}
 
-	err = r.ReconcileCommon(dspa, params)
-	if err != nil {
-		return ctrl.Result{}, err
-	}
+	// Get Prereq Status (DB and ObjStore Ready)
+	dbAvailable := r.isDatabaseAccessible(ctx, dspa, params)
+	objStoreAvailable := r.isObjectStorageAccessible(ctx, dspa, params)
+	dspa_prereqs_ready := (dbAvailable && objStoreAvailable)
```
HumairAK, line 224: > `dspa_prereqs_ready`
 
can you convert this to camelCase

api/v1alpha1/dspipeline_types.go
```diff
@@ -122,6 +122,9 @@ type MlPipelineUI struct {
 type Database struct {
 	*MariaDB    `json:"mariaDB,omitempty"`
 	*ExternalDB `json:"externalDB,omitempty"`
+	// +kubebuilder:default:=false
+	// +kubebuilder:validation:Optional
+	DisableHealthCheck bool `json:"disableHealthCheck"`
```
HumairAK, line 127: can you update this yaml with the addition of these [new fields](https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/config/samples/dspa_all_fields.yaml):

HumairAK:
functionality wise this lgtm, just a few minior changes above

controllers/database.go
```diff
@@ -31,6 +36,52 @@ var dbTemplates = []string{
...
 }
 
+// extract to var for mocking in testing
+var ConnectAndQueryDatabase = func(host, port, username, password, dbname string) bool {
+	connectionString := fmt.Sprintf("%s:%s@tcp(%s:%s)/%s", username, password, host, port, dbname)
+	db, err := sql.Open("mysql", connectionString)
+	if err != nil {
+		return false
+	}
+	defer db.Close()
```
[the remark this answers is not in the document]
gmfrasca: this block (with 'unhandled error/param' from calling `defer db.Close()`) is a common/ubiquitous db connection pattern, it would be more confusing to diverge from the norm here

amadhusu (approved):
/lgtm

PR #154, opened 2023-06-08
Update OCP pipelines pre-req version req.

HumairAK:
As per title.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,9 @@ To get started you will first need to satisfy the following pre-requisites:
 ## Pre-requisites
 1. An OpenShift cluster that is 4.9 or higher.
 2. You will need to be logged into this cluster as [cluster admin] via [oc client].
-3. The OpenShift Cluster must have OpenShift Pipelines 1.7.2 or higher installed. Instructions [here][OCP Pipelines Operator].
+3. The OpenShift Cluster must have OpenShift Pipelines 1.8 or higher installed. We recommend channel pipelines-1.8
+   on OCP 4.10 and pipelines-1.9 or pipelines-1.10 for OCP 4.11, 4.12 and 4.13.
+   Instructions [here][OCP Pipelines Operator].
 4. Based on installation type you will need one of the following:
    1. For Standalone method: You will need to have [Kustomize] version 4.5+ installed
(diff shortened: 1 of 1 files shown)


DharmitD (commented):
/lgtm

shalberd:
what is weird is that the page for 4.10 talked about pipelines-1.7 being the default channel, only since the documentation pages for 4.11 does it correctly mention pipelines-1.8 channel, while mentioning the latest channel as the default one if none is selected.

I would say we can write "use the pipelines-1.8 channel" because of someone write a sucscription yaml and does not supply a channel, they will land it latest by default since 4.11. Whether we link to OCP 4.12 or OCP 4.11 page is not that relevant, with the exception of definitely not linking to the page for OCP 4.10 :-)

README.md
```diff
@@ -30,7 +30,7 @@ To get started you will first need to satisfy the following pre-requisites:
 ## Pre-requisites
 1. An OpenShift cluster that is 4.9 or higher.
 2. You will need to be logged into this cluster as [cluster admin] via [oc client].
-3. The OpenShift Cluster must have OpenShift Pipelines 1.7.2 or higher installed. Instructions [here][OCP Pipelines Operator].
+3. The OpenShift Cluster must have OpenShift Pipelines 1.8 or higher installed. Instructions [here][OCP Pipelines Operator].
```
shalberd, line 33: The OpenShift Cluster must have OpenShift Pipelines 1.8 or higher installed. Use the `pipelines-1.8 `channel.

shalberd: Alternative: We always recommend version 1.8.x for all OCP 4 minor versions

shalberd:
very nice, thank you ... I appreciate it. Looks good to me and this way, it's clearer again to people upstream / in community ODH. E.g. for dashboard folks and the new integrations.

@andrewballantyne We updated the Red Hat Openshift Pipelines operator requirements readme.

HumairAK:
> "Using the Graphical UI now the readme could show the new ODH Dashboard with pipeline capabilities, instead of/in addition of the Kubeflow Pipelines UI" i.e. here https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/README.md#using-the-graphical-ui

@shalberd 

Yes agreed, now that we have odh dashboard ui, it makes sense to update these docs to reflect, let's create a separate issue for this.

PR #196, opened 2023-07-04
Infer the "secure" field in dspa based on scheme

### LINKED ISSUE #157: Infer "secure" field in a dspa if it is not provided.
Currently odh-dashboard offers no option to use "secure: false" connection for object storage connection, so the only way to use an insecure "http" connection for your object storage is to go in the backend and update the dspa manually. We should infer this scheme if it's not provided, this way dashboard does not need to provide a field for this.

Acceptance criteria:
Infer "secure" field in dspa if it is not provided (consider removing it entirely but if this api change results in a lot of overhead go with former option).

amadhusu:
## Description
Infer the "secure" field in dspa based on the scheme which resolves #157 

## How Has This Been Tested?
Manually tested this on Openshift Cluster to ensure removing this field doesn't hinder anything.

## Merge criteria:

### DIFF
--- a/controllers/dspipeline_params.go
+++ b/controllers/dspipeline_params.go
@@ -27,6 +27,7 @@ import (
 	mf "github.com/manifestival/manifestival"
 	dspa "github.com/opendatahub-io/data-science-pipelines-operator/api/v1alpha1"
 	"github.com/opendatahub-io/data-science-pipelines-operator/controllers/config"
+	"github.com/opendatahub-io/data-science-pipelines-operator/controllers/util"
 	v1 "k8s.io/api/core/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -63,10 +64,10 @@ type DBConnection struct {
 type ObjectStorageConnection struct {
 	Bucket            string
 	CredentialsSecret *dspa.S3CredentialSecret
-	Secure            bool
 	Host              string
 	Port              string
 	Scheme            string
+	Secure            *bool
 	Endpoint          string // scheme://host:port
 	AccessKeyID       string
 	SecretAccessKey   string
@@ -235,7 +236,17 @@ func (p *DSPAParams) SetupObjectParams(ctx context.Context, dsp *dspa.DataScienc
 		p.ObjectStorageConnection.Bucket = dsp.Spec.ObjectStorage.ExternalStorage.Bucket
 		p.ObjectStorageConnection.Host = dsp.Spec.ObjectStorage.ExternalStorage.Host
 		p.ObjectStorageConnection.Scheme = dsp.Spec.ObjectStorage.ExternalStorage.Scheme
-		p.ObjectStorageConnection.Secure = dsp.Spec.ObjectStorage.ExternalStorage.Secure
+
+		if dsp.Spec.ObjectStorage.ExternalStorage.Secure == nil {
+			if p.ObjectStorageConnection.Scheme == "https" {
+				p.ObjectStorageConnection.Secure = util.BoolPointer(true)
+			} else {
+				p.ObjectStorageConnection.Secure = util.BoolPointer(false)
+			}
+		} else {
+			p.ObjectStorageConnection.Secure = dsp.Spec.ObjectStorage.ExternalStorage.Secure
+		}
+
 		// Port can be empty, which is fine.
 		p.ObjectStorageConnection.Port = dsp.Spec.ObjectStorage.ExternalStorage.Port
 		customCreds = dsp.Spec.ObjectStorage.ExternalStorage.S3CredentialSecret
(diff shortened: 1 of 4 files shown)


shalberd:
So now that secure is removed here ... how do you handle setting secure to either true or false in i.e. odh dashboard?

config/rbac/role.yaml
```diff
@@ -154,6 +154,16 @@ rules:
   - seldondeployments
   verbs:
   - '*'
+- apiGroups:
```
HumairAK, line 157: you are pulling in a commit already in main, can you rebase?

amadhusu: Ugh didn't see that, but wait I think rebase took Dharmit's commit into picture hence the additional commit.

HumairAK:
@amadhusu can you squash your commits

controllers/dspipeline_params.go
```diff
@@ -235,7 +235,11 @@ func (p *DSPAParams) SetupObjectParams(ctx context.Context, dsp *dspa.DataScienc
 		p.ObjectStorageConnection.Bucket = dsp.Spec.ObjectStorage.ExternalStorage.Bucket
 		p.ObjectStorageConnection.Host = dsp.Spec.ObjectStorage.ExternalStorage.Host
 		p.ObjectStorageConnection.Scheme = dsp.Spec.ObjectStorage.ExternalStorage.Scheme
-		p.ObjectStorageConnection.Secure = dsp.Spec.ObjectStorage.ExternalStorage.Secure
+		if p.ObjectStorageConnection.Scheme == "https" {
+			p.ObjectStorageConnection.Secure = true
+		} else {
+			p.ObjectStorageConnection.Secure = false
```
HumairAK, line 241: Because the field exists in the CR spec, we want this field to take precedence if it's set within the CR, so your conditional needs to account for this (i.e. if `.Secure` is set, then use that value, otherwise infer from .Scheme). 

But this introduces a new issue, which is we do have a default value for `Secure`, which is [here](https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/api/v1alpha1/dspipeline_types.go#L218). 

So we need to remove this default value from the crd spec, and always manually set it.

I believe this will introduce a new problem, which is if we don't set this value the default will become "false", since the `zero` value for bool is `false`. To get around this you will need to convert `Secure` to a type `*bool` and check for `nil` if it's unset.

controllers/dspipeline_params.go
```diff
@@ -265,6 +269,8 @@ func (p *DSPAParams) SetupObjectParams(ctx context.Context, dsp *dspa.DataScienc
 		)
 		p.ObjectStorageConnection.Port = config.MinioPort
 		p.ObjectStorageConnection.Scheme = config.MinioScheme
+		p.ObjectStorageConnection.Secure = config.MinioSecure
```
HumairAK, line 272: See above comment, we'll want to check if `.secure` is set here as well before setting some default.

controllers/dspipeline_params.go
```diff
@@ -105,6 +105,10 @@ func passwordGen(n int) string {
 	return string(b)
 }
 
+func BoolPointer(b bool) *bool {
+	return &b
```
HumairAK, line 109: can you move this to [util.go](https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/controllers/util/util.go)

controllers/dspipeline_params.go
```diff
@@ -235,7 +239,15 @@ func (p *DSPAParams) SetupObjectParams(ctx context.Context, dsp *dspa.DataScienc
...
 		p.ObjectStorageConnection.Scheme = dsp.Spec.ObjectStorage.ExternalStorage.Scheme
-		p.ObjectStorageConnection.Secure = dsp.Spec.ObjectStorage.ExternalStorage.Secure
+
+		if dsp.Spec.ObjectStorage.ExternalStorage.Secure == nil {
+			if p.ObjectStorageConnection.Scheme == "https" {
+				p.ObjectStorageConnection.Secure = BoolPointer(true)
+			} else {
+				p.ObjectStorageConnection.Secure = BoolPointer(false)
+			}
+		}
```
HumairAK, line 249: This won't account for when `secure` _is set_, in which case it will be nil and thus crash the server, can you update it to the following: 

```go
if dsp.Spec.ObjectStorage.ExternalStorage.Secure == nil {
	if p.ObjectStorageConnection.Scheme == "https" {
		p.ObjectStorageConnection.Secure = BoolPointer(true)
	} else {
		p.ObjectStorageConnection.Secure = BoolPointer(false)
	}
} else {
	p.ObjectStorageConnection.Secure = dsp.Spec.ObjectStorage.ExternalStorage.Secure
}
```

controllers/dspipeline_params.go
```diff
@@ -265,6 +277,7 @@ func (p *DSPAParams) SetupObjectParams(ctx context.Context, dsp *dspa.DataScienc
 		)
 		p.ObjectStorageConnection.Port = config.MinioPort
 		p.ObjectStorageConnection.Scheme = config.MinioScheme
+
```
HumairAK, line 280: current implementation will crash when using default minio because `.secure` is never set, and will default to `nil` since we use a pointer now, to avoid this, we need to manually set this value to false for this case: 

```go
p.ObjectStorageConnection.Secure = BoolPointer(false)  // we can assume it will be false since we are deploying it
```

controllers/dspipeline_params.go
amadhusu: That was a big miss....re added that else in the PR.

controllers/dspipeline_params.go
amadhusu: I was considering using a one-liner anonymous function for this but this makes it easily readable so moved this to util as suggested.

PR #223, opened 2023-07-26
Added RequeueAfter to ensure periodic reconciliation

### LINKED ISSUE #173: DSPO will not create/reconcile DSPA if External Object Store secret is missing
If the Secret that a DSPA.Spec.ObjectStorage.ExternalStorage references does not exist, the DSPO will log an error and not attempt to create/apply the DSPA.  This is likely fine, but now the DSPA appears to the DSPO as in-sync and will not Reconcile indefinitely, even if the missing secret/key is created.  This means the DSPA will always be stuck as undeployed until the DSPA CR is edited.

amadhusu:
## Description
Added RequeueAfter which periodically reconciles and fixes #173

## How Has This Been Tested?
Manually tested this by ->
1. Created a DSPA referring to a secret that doesn't exist.
2. Create the secret in question.
3. Monitor the DSPO logs and see that it reconciles and gets DSPA up and running as it should. 

## Merge criteria:

### DIFF
--- a/controllers/dspipeline_controller.go
+++ b/controllers/dspipeline_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
 	mf "github.com/manifestival/manifestival"
@@ -161,10 +162,10 @@ func (r *DSPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	dspa := &dspav1alpha1.DataSciencePipelinesApplication{}
 	err := r.Get(ctx, req.NamespacedName, dspa)
 	if err != nil && apierrs.IsNotFound(err) {
-		log.Info("Stop DSPAParams reconciliation")
+		log.Info("DSPA resource was not found")
 		return ctrl.Result{}, nil
 	} else if err != nil {
-		log.Error(err, "Unable to fetch the DSPAParams")
+		log.Error(err, "Encountered error when fetching DSPA")
 		return ctrl.Result{}, err
 	}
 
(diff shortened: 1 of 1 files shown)


HumairAK:
@amadhusu you'll need to rebase

controllers/dspipeline_controller.go
```diff
@@ -161,7 +163,7 @@ func (r *DSPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	err := r.Get(ctx, req.NamespacedName, dspa)
 	if err != nil && apierrs.IsNotFound(err) {
 		log.Info("Stop DSPAParams reconciliation")
-		return ctrl.Result{}, nil
+		return ctrl.Result{Requeue: true, RequeueAfter: 2 * time.Minute}, nil
```
HumairAK, line 166: this is in the wrong place, the logs are misleading, probably became outdated since the params extraction was moved further down, the logs should just read: 

> "Stop DSPAParams reconciliation"  ---> "DSPA resources was not found."
> "Unable to fetch the DSPAParams" --> "Encountered error when fetching DSPA."

As for where we should requeue, it's here: https://github.com/opendatahub-io/data-science-pipelines-operator/blob/dc86c494ecd551a1aea10d426c8951068d8e8adb/controllers/dspipeline_controller.go#L209

amadhusu: Irrespective of the misleading logs, I assumed based on our previous discussion, this was the right place to place the periodic reconciliation even though ExtractParams made more sense. I made the changes now and changed the log messages as suggested which makes more sense.

controllers/dspipeline_controller.go
```diff
@@ -161,10 +162,10 @@ func (r *DSPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	dspa := &dspav1alpha1.DataSciencePipelinesApplication{}
 	err := r.Get(ctx, req.NamespacedName, dspa)
 	if err != nil && apierrs.IsNotFound(err) {
-		log.Info("Stop DSPAParams reconciliation")
+		log.Info("DSPA resources was not found")
```
HumairAK, line 165:
```suggestion
		log.Info("DSPA resource was not found")
```

amadhusu: :+1:

rimolive:
/lgtm

Tested in a dev cluster and it deployed DSPA after the secret creation

HumairAK:
/lgtm
/approved

Tested and works: 

Verification steps: 
1. Create dspa with external db or storaage configured with a secret that doesn't exist, [example](https://github.com/opendatahub-io/data-science-pipelines-operator/tree/main/config/samples/external-object-storage) (but without the the secret)
2. Confirm the error shows up, and the requeue occurs every 2 min
3. Deploy the secret, confirm the operator moves past this error

Issue #177, opened 2023-06-22
Automated releases for DSP/DSPO


HumairAK:
Today we can manually create a tag release that will auto build the images for this tag, for dsp/dspo, but manual work remains. 

Create an automated workflow via gh actions that, when triggered will do the following: 

Suggestion:
* Cut a new branch (in both dsp/dspo), update the params images with the new tag version to be cut
* cut this new tag version in both dsp/dspo repos 

Check with Distributed workflow community on how they are tackling this issue as they have already formulated ADRs around this.

PR #238, opened 2023-07-31
Tag build image exists

HumairAK:
Furthers #177

This change makes it so we ensure we don't over-write existing tags, and also converts the build logic into a re-usable action.

## Testing instructions
No need to test, you can see it in action here: https://github.com/HumairAK/data-science-pipelines-operator/actions/runs/5716850525 see the attempts, it fails when the tag is present, and succeeds when it is not. Quay reports on all deleted tags as well so we need to verify if these tags are indeed deleted by inspecting their "end time stamp", if they have one it means it's deleted. Furthermore, we don't use `podman image exists` because this will require us to pull the image, this solution instead just queries quay directly and does some `yq` parsing to verify existence of tags.

## Checklist

### DIFF
--- a/.github/actions/build/action.yaml
+++ b/.github/actions/build/action.yaml
@@ -1,0 +1,63 @@
+name: "Set up KinD"
+description: "Step to start and configure KinD cluster"
+inputs:
+  IMAGE_REPO:
+    description: "Quay image repo name."
+    required: true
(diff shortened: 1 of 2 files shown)


DharmitD (approved):
/lgtm
/Approve

Issue #191, opened 2023-06-27
MariaDB failing to start in some namespaces


HumairAK:
The default maria db fails to start in some namespaces

The error in question

```

=> sourcing 20-validate-variables.sh ...
=> sourcing 25-validate-replication-variables.sh ...
=> sourcing 30-base-config.sh ...
---> 12:42:12     Processing basic MySQL configuration files ...
=> sourcing 60-replication-config.sh ...
=> sourcing 70-s2i-config.sh ...
---> 12:42:12     Processing additional arbitrary  MySQL configuration provided by s2i ...
=> sourcing 40-paas.cnf ...
=> sourcing 50-my-tuning.cnf ...
---> 12:42:13     Initializing database ...
---> 12:42:13     Running mysql_install_db ...
mkdir: cannot create directory '/var/lib/mysql/data': Permission denied
Fatal error Can't create database directory '/var/lib/mysql/data'

The latest information about mysql_install_db is available at
https://mariadb.com/kb/en/installing-system-tables-mysql_install_db

```

The error seems hard to reproduce, logging this issue to collect more information.

HumairAK:
On this particular cluster, we found that for some reason the `default` SA in a namespace would get `anyuid` assigned. This resulted in the above issue because MariaDB uses the default SA. The fix in this case was to just assign another SA to the MariaDB deployment. Since we're using the default SA anyway, assigning any SA in this case will work. As a workaround running the following should resolve the issue: 

```bash
# set this to your mariadb ns, where the dspa is deployed (your ds-project if using odh-dashboard)
DSPA_NAMESPACE=...  
oc project ${DSPA_NAMESPACE} 
MARIADB_DEPLOYMENT=$(oc get deployments | grep mariadb | awk '{print $1}')
oc create sa mariadb-sa
oc patch deployment ${MARIADB_DEPLOYMENT} --type='json' -p='[{"op": "add", "path": "/spec/template/spec/serviceAccountName", "value": mariadb-sa}]'
```

PR #245, opened 2023-08-02
Add non-default SAs to MariaDB and Minio

### LINKED ISSUE #204: MariaDB should use non-default SA
To avoid situations like this: https://github.com/opendatahub-io/data-science-pipelines-operator/issues/191 
Where due to cluster configurations, default SA may exhibit unpredictable behaviors, we should opt to use a specifically created SA for mariadb. No special privileges needed for this SA.

While we are at it, let's do this for `minio` as well. The remaining deployments all already have their own SAs.

DharmitD:
Resolves #204 

## Testing instructions

Deploy DSPO and DSPA.
Port forward to the mariaDB port:
```
POD_NAME=$(oc get pods -l app=mariadb-sample -o=jsonpath='{.items[0].metadata.name}')
(base) 
oc port-forward ${POD_NAME} 3306:3306
```
Connect to the port, login inyo mysql
```
mysql -h 127.0.0.1 -P 3306 -u root
```
login into MariaDB
```
mariadb-install-db --user=mysql
```
Make sure you're able to connect successfully.

## Checklist

### DIFF
--- a/config/internal/mariadb/deployment.yaml.tmpl
+++ b/config/internal/mariadb/deployment.yaml.tmpl
@@ -26,6 +26,7 @@ spec:
         component: data-science-pipelines
         dspa: {{.Name}}
     spec:
+      serviceAccountName: ds-pipelines-mariadb-sa-{{.Name}}
       containers:
         - name: mariadb
           image: {{.MariaDB.Image}}
--- a/config/internal/mariadb/mariadb-sa.yaml.tmpl
+++ b/config/internal/mariadb/mariadb-sa.yaml.tmpl
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: mariadb-{{.Name}}
+  name: ds-pipelines-mariadb-sa-{{.Name}}
   namespace: {{.Namespace}}
   labels:
     app: mariadb-{{.Name}}
--- a/config/internal/minio/deployment.yaml.tmpl
+++ b/config/internal/minio/deployment.yaml.tmpl
@@ -22,6 +22,7 @@ spec:
         component: data-science-pipelines
         dspa: {{.Name}}
     spec:
+      serviceAccountName: ds-pipelines-minio-sa-{{.Name}}
       containers:
         - args:
             - server
--- a/config/internal/minio/minio-sa.yaml.tmpl
+++ b/config/internal/minio/minio-sa.yaml.tmpl
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: ds-pipelines-minio-sa-{{.Name}}
+  namespace: {{.Namespace}}
+  labels:
+    app: minio-{{.Name}}
+    component: data-science-pipelines
--- a/controllers/database.go
+++ b/controllers/database.go
@@ -31,8 +31,8 @@ const dbSecret = "mariadb/secret.yaml.tmpl"
 var dbTemplates = []string{
 	"mariadb/deployment.yaml.tmpl",
 	"mariadb/pvc.yaml.tmpl",
-	"mariadb/sa.yaml.tmpl",
 	"mariadb/service.yaml.tmpl",
+	"mariadb/mariadb-sa.yaml.tmpl",
 	dbSecret,
 }
 
--- a/controllers/storage.go
+++ b/controllers/storage.go
@@ -34,6 +34,7 @@ var storageTemplates = []string{
 	"minio/deployment.yaml.tmpl",
 	"minio/pvc.yaml.tmpl",
 	"minio/service.yaml.tmpl",
+	"minio/minio-sa.yaml.tmpl",
 	storageSecret,
 }
 


config/internal/mariadb/mariadb-sa.yaml.tmpl
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: mariadb-sa
```
HumairAK, line 4: can you follow the -{{ .Name }} pattern like all the other resources, same for minio sa

HumairAK:
You created the SA's but you didn't assign the deployment to the SA

config/internal/mariadb/mariadb-sa.yaml.tmpl
DharmitD: Done, updated to have the  -{{ .Name }} pattern.

DharmitD:
> you created the SA's but you didn't assign the deployment to the SA, you do this in the minio/mariadb respective deployment resources, without this it will still use the default SA

My bad, I missed out on pushing the deployment changes earlier. Updated the PR now to have the deployment changes, and tested to make sure deployments use these SAs.

config/internal/minio/minio-sa.yaml.tmpl
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: minio-sa-{{.Name}}
```
gmfrasca, line 4: I think it might be better if we have a `ds-pipelines-` prefix here to avoid confusion about what the SA is intended for

config/internal/mariadb/mariadb-sa.yaml.tmpl
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: mariadb-sa-{{.Name}}
```
gmfrasca, line 4: I think it might be better if we have a `ds-pipelines-` prefix here to avoid confusion about what the SA is intended for

controllers/database.go
```diff
@@ -28,6 +28,7 @@ var dbTemplates = []string{
 	"mariadb/pvc.yaml.tmpl",
 	"mariadb/sa.yaml.tmpl",
```
gmfrasca, line 29: what's the reasoning we don't use this SA anyways?

DharmitD:
> does the non-default (by name) serviceaccount have any special RBAC, similar to i.e. odh-dashboard?

No, the non-default (by name) serviceaccount is not intended to have any special RBAC.

controllers/database.go
DharmitD: You're right, the `sa.yaml.tmpl` isn't being used anywhere. I think the idea suggested by the issue: https://github.com/opendatahub-io/data-science-pipelines-operator/issues/204 to have two SAs, one default SA and one non-default. Issue https://github.com/opendatahub-io/data-science-pipelines-operator/issues/191 further mentions an error encountered in connecting the DB and the fix for it was creating another SA. 

The issue here could be that the original `sa.yaml.tmpl` SA wasn't being referenced in the mariaDB deployment. @HumairAK wdyt?

HumairAK: > https://github.com/opendatahub-io/data-science-pipelines-operator/issues/204 to have two SAs, one default SA and one non-default.

the default SA being referred to in #204 is the SA named `default` that exists in every namespace. By specifying a specific SA to use, we are already using the non-default SA. I didn't know we had an SA already for mariadb not being used, if that's the case let's use this one for mariadb

config/internal/minio/minio-sa.yaml.tmpl
DharmitD: Done, updated to have the ds-pipelines- prefix.

config/internal/mariadb/mariadb-sa.yaml.tmpl
DharmitD: Done, updated to have the ds-pipelines- prefix.

config/internal/mariadb/deployment.yaml.tmpl
```diff
@@ -26,6 +26,7 @@ spec:
         component: data-science-pipelines
         dspa: {{.Name}}
     spec:
+      serviceAccountName: mariadb-sa-{{.Name}}
```
gmfrasca, line 29: gotta update this as well to match the updated SA name ;)

config/internal/minio/deployment.yaml.tmpl
```diff
@@ -22,6 +22,7 @@ spec:
         component: data-science-pipelines
         dspa: {{.Name}}
     spec:
+      serviceAccountName: minio-sa-{{.Name}}
```
gmfrasca, line 25: same here

config/internal/mariadb/deployment.yaml.tmpl
DharmitD: Oops, updated locally and didn't push. Pushed now, should be updated.

config/internal/minio/deployment.yaml.tmpl
DharmitD: Updated.

controllers/database.go
DharmitD: Done, updated to have just one SA for mariaDB, and updated the references in the mariaDB deployment and the controller code to point to this SA.

gmfrasca (approved):

dsp-developers:
A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-245`

PR #267, opened 2023-08-14
Release docs

### LINKED ISSUE #187: Document DSP release steps
Outline an `.md` document on the steps required to create a release for dsp.

HumairAK:
Resolves #187

The workflow doc is [here](https://github.com/HumairAK/data-science-pipelines-operator/blob/release_docs/docs/release/release_workflow.md).

In more detail, the PR adds the following: 

* Release workflow docs for how to release dspo/dsp
* A version compatibility yaml, which is used to create a [compatibility markdown](https://github.com/HumairAK/data-science-pipelines-operator/blob/release_docs/docs/release/compatibility.md) doc
* Python scripts to help with release steps:
  * for creating params.env
  * for creating compatibility doc in markdown
* Params.env images are switched to tags for `main` 

## Testing instructions
One can test the scripts included by following the provided release tools [README.md](https://github.com/HumairAK/data-science-pipelines-operator/blob/a012cc9e1a21c014a3b1c02a69640a2b0989acdc/scripts/release/README.md)

## Checklist


dsp-developers:
A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-267`

scripts/release/version_doc.py
```diff
@@ -0,0 +1,34 @@
+import yaml
+
+
+def table(list_of_dicts):
```
gmfrasca, line 4: this seems pretty fragile.  It's assuming:
1. The compatibility YAML has at least 1 dict item
2. All the dict items have exactly identical keysets

scripts/release/version_doc.py
```diff
@@ -0,0 +1,34 @@
...
+    markdown_table = ""
+    markdown_header = '| ' + ' | '.join(map(str, list_of_dicts[0].keys())) + ' |'
+    markdown_header_separator = '|-----' * len(list_of_dicts[0].keys()) + '|'
+    markdown_table += markdown_header + '\n'
+    markdown_table += markdown_header_separator + '\n'
+    for row in list_of_dicts:
+        markdown_row = ""
+        for key, col in row.items():
```
gmfrasca, line 12: Ordering of items in dicts are not guaranteed, but constructing the table this way assumes each dict's keys are iterated accross in the same order.  We need to intelligently construct the table such that the keys are fetched property

scripts/release/version_doc.py
```diff
@@ -0,0 +1,34 @@
...
+    out_file = args.out_file
+    with open(input_file, 'r') as f:
+        versions = yaml.safe_load(f)
+
+    with open('template/version_doc.md', 'r') as vd:
+        final_md = vd.read()
+
+    table_md = table(versions)
+
+    final_md = final_md.replace('<<TABLE>>', table_md)
```
gmfrasca, line 30: If we're going to use a magic keyword we should be a little more clear here on what `<<TABLE>>` is (rename to something like `<<GENERATED_COMPATIBILITY_TABLE>>`)

scripts/release/release.py
```diff
@@ -0,0 +1,50 @@
+import argparse
+import os
+
+from params import params
+from version_doc import version_doc
+
+
+def env_opts(env: str):
```
gmfrasca, line 8: appears to be unused

scripts/release/params.py
```diff
@@ -0,0 +1,94 @@
...
+    "IMAGES_MLMDGRPC":  "ds-pipelines-metadata-grpc",
+    "IMAGES_MLMDWRITER":  "ds-pipelines-metadata-writer",
+    "IMAGES_DSPO":  "data-science-pipelines-operator",
+}
+
+ARCH = "amd64"
+
+
+def fetch_quay_repo_tag_digest(quay_repo, quay_org, tag):
+    api_url = "https://quay.io/api/v1/repository/{0}/{1}/tag/?specificTag={2}".format(
```
gmfrasca, line 20: f-strings plz

scripts/release/params.py
```diff
@@ -0,0 +1,94 @@
...
+        quay_repo,
+        tag
+    )
+    response = requests.get(api_url).json()
+    tags = response['tags']
+
+    if len(tags) == 0 or 'end_ts' in tags[0]:
+        print("Tag does not exist or was deleted.", file=sys.stderr)
+        exit(1)
+    return tags[0]['manifest_digest']
```
gmfrasca, line 31: we should handle if 'manifest_digest' doesn't exist. (perhaps change to `.get('manifest_digest')`

scripts/release/params.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def fetch_rh_repo_tag_digest(repo, tag):
+    registry = "registry.access.redhat.com"
```
gmfrasca, line 35: magic values like this should be defined in global scope

scripts/release/params.py
gmfrasca: ^above applies to most string construction throughout these scripts, as IMO its much cleaner and maintainable than `.format`

scripts/release/params.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # Fetch QUAY Images
+    for image_env_var in QUAY_REPOS:
+        image_repo = QUAY_REPOS[image_env_var]
+        digest = fetch_quay_repo_tag_digest(image_repo, quay_org, tag)
+        image_repo_with_digest = "{0}:{1}".format(image_repo, digest)
+        images.append("{0}=quay.io/opendatahub/{1}".format(image_env_var, image_repo_with_digest))
+
+    # Fetch RH Registry images
+    repo = "ubi8/ubi-minimal"
```
gmfrasca, line 76: we're repeating the same functionality several times so we probably should have a iteration mechanic here instead.

```
rh_images = [{"env": "IMAGES_CACHE", "tag": ubi_minimal_tag, "repo": "ubi8/ubi-minimal": {"ubi8/ubi-minimal", ...}
for image in rh_images:
    # Do stuff and append images
```

scripts/release/params.py
```diff
@@ -0,0 +1,94 @@
...
+    if not amd_img:
+        print("AMD64 arch image not found for repo {0} and tag {1}".format(repo, tag), file=sys.stderr)
+        exit(1)
+
+    sha_digest = amd_img['image_id']
+
+    return sha_digest
+
+
+def params(args):
```
gmfrasca, line 57: give this a better function name like `generate_paramsenv`

scripts/release/params.py
```diff
@@ -0,0 +1,94 @@
...
+    api_url = "https://catalog.redhat.com/api/containers/v1/repositories/registry/{0}/repository/{1}/tag/{2}".format(
+        registry,
+        repo,
+        tag
+    )
+    response = requests.get(api_url).json()
+
+    amd_img = {}
+    for img in response['data']:
+        if img['architecture'] == 'amd64':
```
gmfrasca, line 45: what if `architecture` doesn't exist or stops existing?

scripts/release/version_doc.py
gmfrasca: it's probably not worth it here for a single string-replace, but if we ever wanted to expand capabilities here a better pattern here would to use a templater like jinja fwiw

config/base/params.env
```diff
@@ -1,12 +1,12 @@
...
+IMAGES_PERSISTENTAGENT=quay.io/opendatahub/ds-pipelines-persistenceagent:latest
+IMAGES_SCHEDULEDWORKFLOW=quay.io/opendatahub/ds-pipelines-scheduledworkflow:latest
+IMAGES_MLMDENVOY=quay.io/opendatahub/ds-pipelines-metadata-envoy:latest
+IMAGES_MLMDGRPC=quay.io/opendatahub/ds-pipelines-metadata-grpc:latest
+IMAGES_MLMDWRITER=quay.io/opendatahub/ds-pipelines-metadata-writer:latest
+IMAGES_DSPO=quay.io/opendatahub/data-science-pipelines-operator:latest
+IMAGES_CACHE=registry.access.redhat.com/ubi8/ubi-minimal:8.8
+IMAGES_MOVERESULTSIMAGE=registry.access.redhat.com/ubi8/ubi-micro:8.8
+IMAGES_MARIADB=registry.redhat.io/rhel8/mariadb-103:1
+IMAGES_OAUTHPROXY=registry.redhat.io/openshift4/ose-oauth-proxy:v4.10
```
shalberd, line 12: for ose-oauth-proxy at least, the idea was to not only have a certain label of the image v4.10, but a set digest, pinned down to a point in time, so really this should always be in digest format. See https://github.com/opendatahub-io/odh-manifests/issues/760 Cause the digest could change with that kind of what I think is called a moving/floating tag.

shalberd: > convert these to digests via automated workflow (the digest that matches with that tag at the time of release)

good point, I do see the main point that a certain and common 4.x version tag is most important for the ODH project as a whole. Now, if all subprojects of ODH worked your way, i.e. determining what lands in odh-manifests in the end at a given time, then the point-in-time resolve of the most current digest to go along with v4.10 or whatever agreed-upon version of ose-oauth / openshift is actually not a bad thing. Up to now, I did not see that Github actions mechanism in i.e. model-mesh or dashboard, which is why the odh-manifests PR exists. Basically, I just want to ensure that the digest-values of ose-oauth-proxy in odh-manifests are the same across all components. How that is best achieved is up for discussion, your way seems very elegant.

@LaVLaS @andrewballantyne @rimolive @VaishnaviHire what do you think?

dsp-developers:
A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-267`

dsp-developers:
A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-267`

dsp-developers:
A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-267`

config/base/params.env
HumairAK: I've updated it to include the sha for oauth proxy in `main` branch, as well as the added an option in the workflow to override any entries, when we automate this portion we'll fetch new digests for all tags, with the exception of oauth proxy which we will `--override` to keep it in sync with odh, if in the feature we can implement a similar workflow in other parts of odh then we can remove this flag, and update digests for oauth proxy with every release if needed.

scripts/release/params.py
HumairAK: That would be an api change on rh registry's part, which would probably break...a lot of things if they did that. But I added a check regardless.

dsp-developers:
A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-267`

scripts/release/params.py
HumairAK: yeah I thought about it but the change to reduce repetition would just add the same number of lines/bloat and make it harder to read (since there are special cases for rh registry like the server, it's not as neat as it was for quay), but I added it anyway

scripts/release/version_doc.py
HumairAK: changed the input yaml to explicitly specify a col spec which we then leverage for indexing rows

scripts/release/version_doc.py
HumairAK: > The compatibility YAML has at least 1 dict item
Added a condition for this 

> All the dict items have exactly identical keysets
it's a table so that's what we're expecting, I just put in a pre-condition for this,  I'd rather avoid spending a lot of time for rigorous validation here as it's just a utility script.

scripts/release/params.py
gmfrasca: you added the check, might as well use the cleaner syntax

```suggestion
        if arch == 'amd64':
```

dsp-developers:
A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-267`

dsp-developers:
A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-267`

gmfrasca (commented):
/lgtm

PR #313, opened 2023-09-01
align example pipeline yaml images

tarilabs:
Some tasks and especially root task in pipeline uses: `quay.io/hukhan/python:alpine3.6`
see:
https://github.com/opendatahub-io/data-science-pipelines-operator/blob/b34fbb8402efd3f2092c7931f683d5b906809961/docs/example_pipelines/condition.yaml#L84

other steps used (dockerhub):
`image: python:alpine3.6`
see:
https://github.com/opendatahub-io/data-science-pipelines-operator/blob/b34fbb8402efd3f2092c7931f683d5b906809961/docs/example_pipelines/condition.yaml#L455

so this changes aligns images _from all tasks, to root task image_.

Might incur in pipeline failure "in the middle" as the tasks using different image might incur in backoff etc.

Example as I followed the quickstart in: https://github.com/opendatahub-io/data-science-pipelines-operator#quickstart

Screenshot 2023-09-01 at 16 11 23

Pipeline definition file as coming in this PR with the changes applied, using the same image as the root task does not brake in the middle:

Screenshot 2023-09-01 at 16 26 44

qed.

Aligns task's images to root task image (`quay.io/hukhan/python:alpine3.6`)

cc @danielezonca @HumairAK @rareddy if any other suggestions?

## Testing instructions
manually verified expected behaviour as documented above.

## Checklist

### DIFF
--- a/docs/example_pipelines/condition.yaml
+++ b/docs/example_pipelines/condition.yaml
@@ -81,7 +81,7 @@ spec:
                     pass
                 with open(output_file, 'w') as f:
                     f.write(_output_serializers[idx](_outputs[idx]))
-          image: quay.io/hukhan/python:alpine3.6
+          image: registry.access.redhat.com/ubi8/python-39
         results:
         - name: Output
           type: string
@@ -91,7 +91,7 @@ spec:
             pipelines.kubeflow.org/cache_enabled: "true"
           annotations:
             pipelines.kubeflow.org/component_spec_digest: '{"name": "Flip coin", "outputs":
-              [{"name": "Output", "type": "String"}], "version": "Flip coin@sha256=2895da7ed0b52f45e6d6a4d8837e4c1ee4ba99e14e5877320155243c67e4610e"}'
+              [{"name": "Output", "type": "String"}], "version": "Flip coin@sha256=32d5bd05b9fa18850505b73d6fb8489cc61f83033306230c8e4da12bdd8890e0"}'
     - name: random-num
       taskSpec:
         steps:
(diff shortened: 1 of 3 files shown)


docs/example_pipelines/condition.yaml
```diff
@@ -452,7 +452,7 @@ spec:
             f.close()
           - $(inputs.params.operand1)
           - $(inputs.params.operand2)
-          image: python:alpine3.6
+          image: quay.io/hukhan/python:alpine3.6
```
gmfrasca, line 455: The intent of this is good, but I think we should stay away from using @HumairAK 's personal quay repo.  Could we instead use the UBI python image (`registry.access.redhat.com/ubi8/python-38`), which is also used elsewhere in code/configs around here?

tarilabs: ff6e701 : applied code review.

Tested manually, and still working as intended:

Screenshot 2023-09-01 at 21 18 54

showing new image as suggested in code review behind the scenes:

Screenshot 2023-09-01 at 21 20 54

also in my understanding registry.access.redhat.com does supports unauthenticated access which is paramount being this quickstart upstream/OSS/community -oriented.

Thank you for your guidance @gmfrasca !

gmfrasca (commented):
/lgtm

gmfrasca (commented):
/lgtm

strangiato:
ubi8/python-38 is depreciated and should probably not be used.

https://catalog.redhat.com/software/containers/ubi8/python-38/5dde9cacbed8bd164a0af24a

image

tarilabs:
ace1992 accounts for https://github.com/opendatahub-io/data-science-pipelines-operator/pull/313#issuecomment-1707357976

working as expected (again):

Screenshot 2023-09-06 at 08 23 06

Screenshot 2023-09-06 at 08 23 13

PR #533, opened 2024-01-04
Fix mismatched requeue var name  in DSPO ConfigMap

gmfrasca:
The variable name, as provided to Kustomize, is simply REQUEUE_TIME not DSPO_REQUEUE_TIME.  Currently the config will not expand this value because of the mismatch

## Testing instructions
Deploy DSPO using changeset.  Verify dspo-config ConfigMap now has a proper DSPO.RequeueTime value (expected: `2m`, not un-expanded/literal `$(DSPO_REQUEUE_TIME)`)

## Checklist

### DIFF
--- a/config/configmaps/files/config.yaml
+++ b/config/configmaps/files/config.yaml
@@ -28,4 +28,4 @@ DSPO:
       ConnectionTimeout: $(DSPO_HEALTHCHECK_DATABASE_CONNECTIONTIMEOUT)
     ObjectStore:
       ConnectionTimeout: $(DSPO_HEALTHCHECK_OBJECTSTORE_CONNECTIONTIMEOUT)
-  RequeueTime: $(DSPO_REQUEUE_TIME)
+  RequeueTime: $(REQUEUE_TIME)


dsp-developers:
A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-533`

```bash
cd $(mktemp -d)
git fetch origin pull/533/head
git checkout -b pullrequest 3a50314c40dbefd351f7378853eb0184ce0ea750
make deploy IMG="quay.io/opendatahub/data-science-pipelines-operator:pr-533"
```

config/configmaps/files/config.yaml
```diff
@@ -28,4 +28,4 @@ DSPO:
       ConnectionTimeout: $(DSPO_HEALTHCHECK_DATABASE_CONNECTIONTIMEOUT)
     ObjectStore:
       ConnectionTimeout: $(DSPO_HEALTHCHECK_OBJECTSTORE_CONNECTIONTIMEOUT)
-  RequeueTime: $(DSPO_REQUEUE_TIME)
+  RequeueTime: $(REQUEUE_TIME)
```
DharmitD, line 31: Shouldn't it be `DSPO_REQUEUE_TIME` itself? The field path in the kustomize yaml points to `DSPO_REQUEUE_TIME`: https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/config/base/kustomization.yaml#L135

DharmitD: Actually nvm, let'd just stick to `REQUEUE_TIME` everywhere. 

I think this would've worked if the variable in `params.env` were `REQUEUE_TIME` instead of `DSPO_REQUEUE_TIME`: https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/config/base/params.env

Could you update it to be `REQUEUE_TIME` here and in the kustomization.yaml too?

gmfrasca: so i checked the full codebase.... i wonder if we should converge on `DSPO_REQUEUETIME` such that it would theoretically also be picked up by viper?  it wouldn't _really_ matter since the value would be reflected in both envvars and config, but it would remove some confusion

DharmitD: Sounds good, let's converge on `DSPO_REQUEUETIME`

HumairAK:
already implemented elsewhere

PR #556, opened 2024-01-24
Updating NW Policy to have notebook pod connect to DSP APIServer

DharmitD:
Resolves [RHOAIENG-1689](https://issues.redhat.com/browse/RHOAIENG-1689)

Updating the DSPA network policy to have notebook pod connect to DSP APIServer, so that the notebook pods could connect to the DSP API Server without OAuth token interference.

## Testing instructions

Deploy DSPO and a DSPA instance based on these changes.

## Checklist

### DIFF
--- a/config/internal/common/tekton/policy.yaml.tmpl
+++ b/config/internal/common/tekton/policy.yaml.tmpl
@@ -64,6 +64,12 @@ spec:
             matchLabels:
               app: ds-pipeline-metadata-writer-{{.Name}}
               component: data-science-pipelines
+        - podSelector:
+            matchLabels:
+              opendatahub.io/dashboard: 'true'
+          namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: {{.Namespace}}
       ports:
         - protocol: TCP
           port: 8888


dsp-developers:
  A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-556`

config/internal/common/tekton/policy.yaml.tmpl
```diff
@@ -64,6 +64,12 @@ spec:
             matchLabels:
               app: ds-pipeline-metadata-writer-{{.Name}}
               component: data-science-pipelines
+        - podSelector:
+            matchLabels:
+              opendatahub.io/dashboard: 'true'
```
gregsheremeta, line 69: Posting here for posterity. Before he posted the PR, I asked Dharmit, Is this the label that @harshad16 recommended to use?

Turns out this is currently the only static, reliable label on every notebook pod. However, I don't like that it's named "dashboard" and I'm not sure if it will always stay there. And I believe @DharmitD has opened a Jira to ask for a better label (is that right @DharmitD )?

So I wonder if we'd be better off somehow templating the label name we're matching as well, to make it less brittle.

What do others think?

DharmitD: Yes, this is the only static label on notebook pods currently, here's the task we've created requesting for an exclusive static label for notebooks:  https://github.com/opendatahub-io/notebooks/issues/387

I agree that "dashboard" name on the label doesn't sound right, but I'm out of ideas on how we could template our network policy to grab the notebook pod name label, or any other dynamic notebook labels. @harshad16 or anyone else might have some ideas?

gregsheremeta: Nah, there's no way that we could template in a dynamic notebook label at DSPA creation time, which is when the NetworkPolicy gets inserted. A possible option would be to have something on the notebook side ("notebook controller"? I don't know the architecture over there) be the thing that lays down a NetworkPolicy update at notebook pod creation time.

On a separate topic ... how secure is it to allow something to bypass our oauth proxy with a simple label (granted, the pod being labeled has to be in the correct namespace too...) ?

harshad16: Here is my chain of thought: 
Based on the discussion to which I was invited, it was just been asked which labels are currently available on the workbench that is standard across all workbenches.
Currently, `opendatahub.io/dashboard: 'true'` is the only label that is added to all workbenches.

Going forward, we would have to create a label that exclusively attaches to all the workbenches.
that being said: pointing out, we have the requirement, that existing notebooks are not to be hindered,
so our logic of applying a new label would only apply to new workbenches or restarted ones, not to the existing Notebook in the cluster.
Based on this, if we would like to address all workbenches, then the current label is the only option, if that's not the case, we can create a new label for new workbenches.

gregsheremeta: makes sense. I'd still rather have a new label, and maybe we can say something like "you need to recreate your notebook if you want to be able to seamlessly connect to DSP"

gregsheremeta: ok so since we know the label will change, I'm looking for another opinion on whether we should leave it as a hardcoded string here, or if we should template. I'm personally ok with the hardcode, but I'm worried that I'm not thinking through upgrade scenarios.

gregsheremeta:
mostly lgtm at this point, but I want someone who knows more about NewtworkPolicy and overall kubernetes security to review this. I want to know if there are any risks to bypassing oauth this way. Could an attacker easily bypass DSP authentication by adding a label to the attacking pod? (I suppose if someone has priveleges to set a label on a pod in the namespace, they could probably directly access the DSP pod anyway?)

gregsheremeta:
@strangiato wouldn't mind your opinion on this. You seem knowledgeable about stuff like this.

PR #575, opened 2024-02-20
V1.6.x rhoaieng 1690 and 2385 2

HumairAK:
Resolves for v1.6: 

* https://issues.redhat.com/browse/RHOAIENG-1690
* https://issues.redhat.com/browse/RHOAIENG-2099

This change allows the DSPO to perform heatlh checks against a tls
secured database. If the database is behind a self-signed cert, end user
can provide a ca-bundle either as a global cert (configmap named
"odh-trust-bundle" or via the ".caBundle" config option via the DSPA.

This change also exposes the "ExtraParams" field for DSP, meaning users
can now add ny DSN parameters when configuring a DB connection for DSP.
These params are utilized in the same manner by the DSPO when conducting
the health check to keep the behavior consistent.

## Testing instructions

Deploy a secure/tls enabled s3/mariadb behind a self-signed cert (for example in a self-signed ocp cluster), provide these configs to the DSAP as external configs.  Ensure they only accept tls based connections.

Deploy DSPA configured to leverage these connections via external db/object store connections. 

Confirm DSPO is able conduct successfull health checks (if it deploys the DSPA pods then health checks worked), if it didn't, log the error here. 

Confirm DSPA comes up successfully, in this change DSPO willl udpate the DSP server's configs to enable "tls" for db connections. Confirm this behavior by running a successful pipeline. 

## Checklist


dsp-developers:
  A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-575`

amadhusu (commented):
I was able to test and verify the first two test cases but not the last one with the new field "customExtraParams".

> Confirm setting "customExtraParams" in the DSPA's ".spec.database.customExtraParams". Provide wrong ca bundle, confirm it fails to deploy dspa, then set ".spec.database.customExtraParams" to  {"tls":"true"}, it should work.

amadhusu (approved):
Tested the `customExtraParams` after Humair's suggestion about the format. It works as expected.

controllers/dspipeline_controller.go
```diff
@@ -516,6 +516,38 @@ func (r *DSPAReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&rbacv1.Role{}).
 		Owns(&rbacv1.RoleBinding{}).
 		Owns(&routev1.Route{}).
+		// Watch for global ca bundle, if one is added to this namespace
+		// we need to reconcile on all the dspa's in this namespace
+		// so they may mount this cert in the appropriate containers
+		Watches(&source.Kind{Type: &corev1.ConfigMap{}},
```
gregsheremeta, line 522: watching ConfigMap gobbles up memory doesn't it? Do we need to check for that and mitigate, possibly by increasing limits and or requests?

controllers/dspipeline_params.go
```diff
@@ -149,6 +151,9 @@ func (p *DSPAParams) SetupDBParams(ctx context.Context, dsp *dspa.DataSciencePip
 		p.DBConnection.Port = dsp.Spec.Database.ExternalDB.Port
 		p.DBConnection.Username = dsp.Spec.Database.ExternalDB.Username
 		p.DBConnection.DBName = dsp.Spec.Database.ExternalDB.DBName
+		// Assume default external connection is tls enabled
+		// user can override this via CustomExtraParams field
+		p.DBConnection.ExtraParams = fmt.Sprintf(config.DBDefaultExtraParams, true)
```
gregsheremeta, line 156: The extra params should be a struct that gets serialized to json when it needs to be. I guess it needs to by dynamic though, because you don't want to chase the mysql API. Hmm. Sprintf'ing the 'true' or 'false' specifically for tls feels weird when the field is named "DBDefaultExtraParams" and not "tlsEnabled". Not sure what the solution is but wanted to call out that it feels weird to me.

controllers/dspipeline_params.go
```diff
@@ -468,16 +489,39 @@ func (p *DSPAParams) ExtractParams(ctx context.Context, dsp *dspa.DataSciencePip
...
+					log.Info(fmt.Sprintf("Encountered error when attempting to fetch ConfigMap: [%s], Error: %v", dspaCaBundleCfgName, dspaCACfgErr))
+					return dspaCACfgErr
+				}
+				p.APICustomPemCerts = []byte(dspaProvidedCABundle)
+			}
+		} else if err != nil {
+			log.Info(fmt.Sprintf("Encountered error when attempting to fetch ConfigMap: [%s], Error: %v", globalCABundleCFGMapKey, err))
+			return err
+		} else {
+			// Found a global cert, consume this cert, takes precedence over "cABundle" provided via DSPA
```
gregsheremeta, line 517: local usually overrides global ... pinged you in slack about it

gregsheremeta (commented):
i have some thoughts and nitpicks ... all of which we can deal with in a follow up

/lgtm

config/internal/apiserver/server-config.yaml.tmpl
```diff
@@ -0,0 +1,18 @@
...
+        component: data-science-pipelines
+data:
+    config.json: |
+      {
+        "DBConfig": {
+          "DriverName": "mysql",
+          "ConMaxLifeTime": "120s",
+          "ExtraParams": {{ .DBConnection.ExtraParams }}
+        },
+        "InitConnectionTimeout": "6m"
```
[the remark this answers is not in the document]
HumairAK: it's the default, kfp requires this field for this config

controllers/dspipeline_controller.go
HumairAK: We added a similar watcher for a pod, and didn't notice a major uptick (a lot of this is internally cached and heavily optimized in k8s I believe), and pod events work queue is likely to be insanely longer than the configmap events

HumairAK: buut if it does cause issues, we will probably catch it in our perf testing and adjust accordingly

controllers/dspipeline_params.go
HumairAK: I actually did it that way at first, i.e. Extraparms was of type `map[string]string` but in our configmap template we need to render it as a string, the thing is we need to pass it in as a string anyways to render it in our go template configmap, I was torn on it, this method resulted in a lot less lines of code so I did it this way instead

dsp-developers:
  A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-575`

PR #590, opened 2024-03-06
chore: remove unneeded image paths in source build wf

HumairAK:


.github/workflows/build-tags.yml
```diff
@@ -66,14 +66,10 @@ env:
...
   IMAGE_REPO_SERVER: ds-pipelines-api-server
   IMAGE_REPO_UI: ds-pipelines-frontend
-  IMAGE_REPO_CACHE: ds-pipelines-cacheserver
   IMAGE_REPO_PA: ds-pipelines-persistenceagent
   IMAGE_REPO_SWF: ds-pipelines-scheduledworkflow
-  IMAGE_REPO_VC: ds-pipelines-viewercontroller
-  IMAGE_REPO_ARTIFACT: ds-pipelines-artifact-manager
   IMAGE_REPO_LAUNCHER: ds-pipelines-launcher
   IMAGE_REPO_DRIVER: ds-pipelines-driver
-  IMAGE_REPO_MLMD_WRITER: ds-pipelines-metadata-writer
```
gmfrasca, line 76: don't we still need this for v1 compatibility?

.github/workflows/build-tags.yml
```diff
@@ -66,14 +66,10 @@ env:
   IMAGE_REPO_DSPO: data-science-pipelines-operator
   IMAGE_REPO_SERVER: ds-pipelines-api-server
   IMAGE_REPO_UI: ds-pipelines-frontend
-  IMAGE_REPO_CACHE: ds-pipelines-cacheserver
   IMAGE_REPO_PA: ds-pipelines-persistenceagent
   IMAGE_REPO_SWF: ds-pipelines-scheduledworkflow
-  IMAGE_REPO_VC: ds-pipelines-viewercontroller
-  IMAGE_REPO_ARTIFACT: ds-pipelines-artifact-manager
```
gmfrasca, line 73: same as below

.github/workflows/build-tags.yml
HumairAK: this build references only one repo, for v1 compatibility we'd need to update the entire wf to support multiple repos, right now we don't build these for v1 support from the dsp repo anyway so it doesn't make sense to keep them

.github/workflows/build-tags.yml
HumairAK: https://github.com/opendatahub-io/data-science-pipelines-operator/pull/590#discussion_r1515120404

DharmitD (commented):
/lgtm

.github/workflows/build-tags.yml
HumairAK: actually that's irrelevant, we didn't use this at all

.github/workflows/build-tags.yml
HumairAK: we did not use these

gmfrasca (approved):

PR #617, opened 2024-03-25
chore: add ssl env to functest

HumairAK:
Have functest make command set the custom SSL_CERT_FILE path to leverage dummy cabundles


gmfrasca (commented):
/lgtm

PR #702, opened 2024-09-17
Move kind-integration.sh content to tests.sh and target environment for testing

diegolovison:
Break down the solution for #[RHOAIENG-10212](https://issues.redhat.com//browse/RHOAIENG-10212) into smaller PRs to facilitate review.

The goal of test.sh is to be the starting point for allowing a test to be running to the following target: kind, fresh OpenShift, an OpenShift cluster with RHOAI installed.

### How to review:
* The content of `kind-integration.sh` was copied to `test.sh`
* Functions were created into `test.sh` 
* A new way to call the script was created `sh tests.sh --kind`
* In the next PR, we are introducing the `--openshift` parameter and implementation

### Why this should be merged
* No manual testing is required
* Defined a new way to call the script that will be the base for `--openshift`. Having the base well defined and agreement from the team we can move forward and avoid rework. There are multiple ways to create a bash script and we should use one that the team is more comfortable with.

### Questions
*  Explain the targets and the difference: target: kind, fresh OpenShift, an OpenShift cluster with RHOAI installed.
* * kind: The script will create all namespaces and set up all required dependencies. Example: routes, CRDs, 
* * Fresh OpenShift: The script will ( next PR ) create all namespaces required but will skip Routes ( OpenShift native ). It will also install the required dependencies.
* * OpenShift cluster with RHOAI installed: It won't install namespaces, routes, or CRDs. It will install only required depedencies.
* * * All the 3 options above are setup related things, after the setup, the test will run. The test code is the same for all the 3 install options above.

## Testing instructions
If GHA action passes, all good. No manual testing is required.

## Checklist


HumairAK:
/approve

will let @rimolive lgtm it

.github/scripts/tests/tests.sh
```diff
@@ -1,5 +1,220 @@
-#!/usr/bin/env bash
+#!/bin/bash
+# This is script is defined as the following:
+# 1 - We declare the required environment variables
+# 2 - Has the functions defined
+# 3 - Setup the environment and run the tests by using the appropriated functions
 
-set -ex
+set -e
```
rimolive, line 7: My suggestion: `set -euo pipefail`

From bash man page:
* -e: Exit immediately if a pipeline returns a non-zero status
* -u: Treat unset variables and parameters other than the special parameters ‘@’ or ‘*’, or array variables subscripted with ‘@’ or ‘*’, as an error when performing parameter expansion
* -o pipefail: If set, the return value of a pipeline is the value of the last (rightmost) command to exit with a non-zero status, or zero if all commands in the pipeline exit successfully.

diegolovison: Done!

PR #711, opened 2024-09-30
Add support for using route or service as target endpoints in the test suite

diegolovison:
Resolves partially [RHOAIENG-10212](https://issues.redhat.com//browse/RHOAIENG-10212)

This PR introduces enhancements to the test suite by allowing the use of either a route or a Kubernetes service as the target endpoint for interacting with the Data Science Pipelines Operator (DSPO).

1. **New `--endpointType` Flag**:
   - A new `--endpointType` flag is introduced in the test suite configuration, which specifies whether the test should use a Kubernetes service (`service`) or an OpenShift route (`route`) as the target endpoint for DSPO.
   
2. **Infrastructure Cleanup Option**:
   - The `--cleanup` flag is renamed to `--clean-infra`. This option allows automated cleanup of infrastructure before rerunning tests.

3. **Modifications in Test Scripts**:
   - Updated the shell scripts and Go test files to accommodate the new endpoint selection mechanism.
   - The HTTP client in the tests is now initialized with custom transport to support token-based authentication when necessary.
   - Various test functions now rely on the `Clientmgr.httpClient` to perform HTTP requests, decoupling from the standard Go `http` package.

4. **New Test Resources**:
   - Two new DSPA (Data Science Pipelines Application) YAML resources (`dspa.yaml` and `dspa-external.yaml`) are added for testing RHOAI.

5. **Test Case Improvements**:
   - Improved the pod listing logic by retrieving only running pods, ensuring the test suite can be rerun multiple times without issues.

## Testing instructions
No manual testing required.
* `kind` it is done by GHA
* `rhoai`: a new PR will be added soon to allow passing: `-endpointType` via `test.sh`. If you wish to test RHOAI ( please call me first ) do the following:
** Modify `Makefile` and append `-endpointType=route` in the `integrationtest`
** Checkout https://github.com/diegolovison/ods-ci/tree/e2e
** From your local machine:
*** `oc login ....`
*** `sh run_robot_test.sh --set-urls-variables true --extra-robot-args '-L DEBUG -i ODS-2632' --skip-oclogin`

## Checklist

### DIFF
--- a/tests/dspa_v2_test.go
+++ b/tests/dspa_v2_test.go
@@ -52,8 +52,10 @@ func (suite *IntegrationTestSuite) TestDSPADeployment() {
 	suite.T().Run("with default MariaDB and Minio", func(t *testing.T) {
 		t.Run(fmt.Sprintf("should have %d pods", podCount), func(t *testing.T) {
 			podList := &corev1.PodList{}
+			// retrieve the running pods only, to allow for multiple reruns of  the test suite
 			listOpts := []client.ListOption{
 				client.InNamespace(suite.DSPANamespace),
+				client.MatchingFields{"status.phase": string(corev1.PodRunning)},
 			}
 			err := suite.Clientmgr.k8sClient.List(suite.Ctx, podList, listOpts...)
 			require.NoError(t, err)
--- a/tests/resources/dspa-external.yaml
+++ b/tests/resources/dspa-external.yaml
@@ -1,0 +1,41 @@
+apiVersion: datasciencepipelinesapplications.opendatahub.io/v1alpha1
+kind: DataSciencePipelinesApplication
+metadata:
+  name: dspa-ext
+spec:
+  dspVersion: v2
+  podToPodTLS: true
+  apiServer:
+    deploy: true
+    enableOauth: true
+    enableSamplePipeline: true
+    cABundle:
+      configMapName: root-ca
+      configMapKey: public.crt
+  scheduledWorkflow:
+    deploy: true
+  persistenceAgent:
+    deploy: true
+  mlmd:
+    deploy: true
+  database:
+    customExtraParams: '{"tls":"true"}'
+    externalDB:
+      host: mariadb.test-mariadb.svc.cluster.local
+      port: "3306"
+      username: mlpipeline
+      pipelineDBName: mlpipeline
+      passwordSecret:
+        name: ds-pipeline-db-test
+        key: password
+  objectStorage:
+    externalStorage:
+      bucket: mlpipeline
+      host: minio.test-minio.svc.cluster.local
+      port: "9000"
+      region: us-east-2
+      s3CredentialsSecret:
+        accessKey: accesskey
+        secretKey: secretkey
+        secretName: minio
+      scheme: https
--- a/tests/suite_test.go
+++ b/tests/suite_test.go
@@ -20,8 +20,12 @@ package integration
 
 import (
 	"context"
+	"crypto/tls"
 	"flag"
 	"fmt"
+	routev1 "github.com/openshift/api/route/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"log"
 	"testing"
 	"time"
 
@@ -40,6 +44,7 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"net/http"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
@@ -62,6 +67,7 @@ var (
 	PortforwardLocalPort int
 	DSPA                 *dspav1alpha1.DataSciencePipelinesApplication
 	forwarderResult      *forwarder.Result
+	endpointType         string
 )
 
 var (
@@ -82,12 +88,14 @@ const (
 	DefaultSkipDeploy           = false
 	DefaultSkipCleanup          = false
 	DefaultDSPAPath             = ""
+	DefaultEndpointType         = "service"
 )
 
 type ClientManager struct {
-	k8sClient client.Client
-	mfsClient mf.Client
-	mfopts    mf.Option
+	k8sClient  client.Client
+	mfsClient  mf.Client
+	mfopts     mf.Option
+	httpClient http.Client
 }
 
 type IntegrationTestSuite struct {
@@ -100,6 +108,11 @@ type IntegrationTestSuite struct {
 
 type testLogWriter struct {
 	t *testing.T
+}
+
+type customTransport struct {
+	Transport http.RoundTripper
+	Token     string
 }
 
 func (w *testLogWriter) Write(p []byte) (n int, err error) {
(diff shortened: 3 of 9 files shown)


tests/dspa_v2_test.go
```diff
@@ -52,8 +52,10 @@ func (suite *IntegrationTestSuite) TestDSPADeployment() {
 	suite.T().Run("with default MariaDB and Minio", func(t *testing.T) {
 		t.Run(fmt.Sprintf("should have %d pods", podCount), func(t *testing.T) {
 			podList := &corev1.PodList{}
+			// we retrieve the running pods only because I can rerun the test suite multiple times
```
HumairAK, line 55:
```suggestion
			// retrieve the running pods only, to allow for multiple reruns of  the test suite
```

tests/suite_test.go
```diff
@@ -176,23 +195,83 @@ func (suite *IntegrationTestSuite) SetupSuite() {
...
+			if err == nil {
+				ok = true
+				break
+			}
+			count++
+			time.Sleep(1 * time.Second)
+		}
+		if !ok {
+			log.Fatal("failed to retrieve router")
+		}
```
HumairAK, line 243: dspa has already readied here: https://github.com/opendatahub-io/data-science-pipelines-operator/pull/711/files#diff-96e66e1ccb0083ac5a2b7fc37cb002803d242ea67c1981a5d09ef8ee2668354cR194

so the route should already have been created, so we shouldn't need to have retry attempts here, it should just work the first time.

tests/suite_test.go
```diff
@@ -176,23 +195,83 @@ func (suite *IntegrationTestSuite) SetupSuite() {
...
+			if response.StatusCode == 200 {
+				ok = true
+				break
+			}
+			count++
+			time.Sleep(1 * time.Second)
+		}
+		if !ok {
+			log.Fatal("Pods didn't started properly")
+		}
```
HumairAK, line 270: this is already covered with this test here: https://github.com/opendatahub-io/data-science-pipelines-operator/pull/711/files#diff-d430921125862c61b7f677a4fa66220726faac5ef8827551fb3accd3ee5c1bd1R32


pod readiness check already queries the healthz endpoint

tests/suite_test.go
```diff
@@ -176,23 +195,83 @@ func (suite *IntegrationTestSuite) SetupSuite() {
...
+		APIServerURL = fmt.Sprintf("https://%s", route.Status.Ingress[0].Host)
+
+		loggr.Info(fmt.Sprintf("The Api Server URL was retrieved from the Route: %s", APIServerURL))
+
+		suite.Clientmgr.httpClient.Transport = &customTransport{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+			},
+			Token: cfg.BearerToken,
+		}
```
HumairAK, line 254: maybe in the future we can use the dsp custom cert to pass here
that would ensure our dsp custom cert is working properly

HumairAK (changes requested):

tests/suite_test.go
diegolovison: No. It is not like that. It is async and the route can take time to be up and running. I implemented the code after seeing failures in the test suite.

tests/suite_test.go
diegolovison: What you wrote is the desirable state but it is not working like that. The code was implemented after seeing failures. We can address and fix it later. The more we really in the readiness check the better.

tests/suite_test.go
diegolovison: Exactly. There is a JIRA ( already opened ) to address that

tests/suite_test.go
```diff
@@ -176,23 +195,83 @@ func (suite *IntegrationTestSuite) SetupSuite() {
...
+
+		// Define the namespaced name
+		key := types.NamespacedName{
+			Namespace: DSPANamespace,
+			Name:      "ds-pipeline-" + suite.DSPA.Name,
+		}
+
+		route := &routev1.Route{}
+
+		// Retrieve the router
```
hbelmiro, line 229:
```suggestion
		// Retrieve the route
```

tests/suite_test.go
```diff
@@ -209,3 +288,11 @@ func (suite *IntegrationTestSuite) TearDownSuite() {
 func TestIntegrationTestSuite(t *testing.T) {
 	suite.Run(t, new(IntegrationTestSuite))
 }
+
+func (t *customTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Add the Authorization header globally
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", t.Token))
```
hbelmiro, line 294: Just to be sure... Did you replace the standard HTTP client because you needed to set this header?

tests/dspa_v2_test.go
diegolovison: Resolved

tests/suite_test.go
diegolovison: Resolved

tests/suite_test.go
```diff
@@ -176,23 +195,83 @@ func (suite *IntegrationTestSuite) SetupSuite() {
...
+			err := clientmgr.k8sClient.Get(context.TODO(), key, route)
+			if err == nil {
+				ok = true
+				break
+			}
+			count++
+			time.Sleep(1 * time.Second)
+		}
+		if !ok {
+			log.Fatal("failed to retrieve router")
```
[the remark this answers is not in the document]
diegolovison: Resolved

tests/suite_test.go
diegolovison: The header for authentication on OpenShift and the Transport. Now, we are ignoring the cert but we are going to implement it soon

```
		suite.Clientmgr.httpClient.Transport = &customTransport{
			Transport: &http.Transport{
				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
			},
			Token: cfg.BearerToken,
		}
```

hbelmiro (commented):
/lgtm

tests/suite_test.go
HumairAK: okay, in that case my suggestion is to re-use the waitFor logic that already exists instead of re-implementing it here, see: https://github.com/opendatahub-io/data-science-pipelines-operator/blob/54c54dd248a6c03ddac6760f4c86c7f0d9b40ddb/tests/util/resources.go#L126

HumairAK: you could also just adjust the condition for what a successful deployDSPA is here https://github.com/opendatahub-io/data-science-pipelines-operator/pull/711/files#diff-96e66e1ccb0083ac5a2b7fc37cb002803d242ea67c1981a5d09ef8ee2668354cR189 

and consider the Ready condition as well

tests/resources/dspa-external.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: datasciencepipelinesapplications.opendatahub.io/v1alpha1
+kind: DataSciencePipelinesApplication
+metadata:
+  name: dspa-ext
```
HumairAK, line 4: how is this different from dspa-ext in the /tests/resources/ folder? are we duplicated files ? can we just use `kustomization.yaml` to reference this instead of duplicating it?

HumairAK:
@diegolovison will provide a follow up pr to address the changes 


/lgtm
/approve

PR #713, opened 2024-10-03
Replace boto3 with numpy

diegolovison:
Resolves https://issues.redhat.com/browse/RHOAIENG-13915

### Explanation of `pip download` and `pip install`

- **pip download**: This command fetches packages and their dependencies without installing them. It is crucial that the version of Python for which the packages are downloaded matches the version in the target runtime environment. If you run `pip download` in an environment with Python 3.9, but your target system uses Python 3.10, there could be incompatibilities.
  
 - **pip install**: This installs the downloaded packages into the runtime environment. Without a Dockerfile, installation can be influenced by the host environment’s Python version, leading to mismatches between the development, testing, and production environments.

1. **Replacement of boto3 with numpy**. It has low dependency artifacts.
   
2. **Environment-Agnostic Package Management**:
   - Download packages for multiple Python versions (`3.9`, `3.10`, `3.11`, `3.12`), ensuring compatibility with different runtime environments.
   
3. **Dockerfile and Shell Script Updates**:
   - This makes the process more reliable by standardizing the environment, making it easier to manage dependencies for pipelines.

### Dockerfile
By using Docker, we create a containerized environment where the Python version is explicitly defined, making the process environment agnostic. The Docker container ensures that no matter where the code runs, the same Python version and packages are used, avoiding discrepancies between development machines, CI/CD pipelines, and production.

This solves the problems with pip download and pip install ( as described above )

## Testing instructions
Not needed. It was done by GHA

## Checklist

### DIFF
--- a/.github/scripts/python_package_upload/Dockerfile
+++ b/.github/scripts/python_package_upload/Dockerfile
@@ -1,0 +1,15 @@
+FROM docker.io/python:3.9
+
+# Set working directory
+WORKDIR /app
+
+# Copy the script into the container
+COPY package_upload.sh /app/package_upload.sh
+
+# Make sure the script is executable
+RUN chmod +x /app/package_upload.sh
+
+# Store the files in a folder
+VOLUME /app/packages
+
+ENTRYPOINT ["/app/package_upload.sh"]
--- a/tests/resources/test-pipeline-with-custom-pip-server.py
+++ b/tests/resources/test-pipeline-with-custom-pip-server.py
@@ -3,10 +3,12 @@ from kfp import dsl, compiler
 # Edited the compiled version manually, to remove the --trusted-host flag
 # this is so we can test for tls certs validations when launcher installs packages
 @dsl.component(base_image="quay.io/opendatahub/ds-pipelines-ci-executor-image:v1.0",
-packages_to_install=['boto3'],
-pip_index_urls=['https://nginx-service.test-pypiserver.svc.cluster.local:443/simple/'])
+packages_to_install=['numpy'],
+pip_index_urls=['https://nginx-service.test-pypiserver.svc.cluster.local/simple/'],
+pip_trusted_hosts=[])
 def say_hello() -> str:
-    hello_text = 'Hello!'
+    import numpy as np
+    hello_text = f'Numpy version: {np.__version__}'
     print(hello_text)
     return hello_text
 
@@ -18,4 +20,4 @@ def hello_pipeline() -> str:
 
 
 if __name__ == '__main__':
-    compiler.Compiler().compile(hello_pipeline, __file__ + '.yaml')
+    compiler.Compiler().compile(hello_pipeline, __file__.replace('.py', '-run.yaml'))
--- a/.github/scripts/python_package_upload/package_upload.sh
+++ b/.github/scripts/python_package_upload/package_upload.sh
@@ -1,33 +1,12 @@
-#!/usr/bin/env bash
+#!/bin/bash
 
 set -ex
 
-kfp_directory=kfp
-boto_directory=boto3
-
-mkdir -p "$kfp_directory"
-mkdir -p "$boto_directory"
-
-# Download kfp package
-pip download kfp==2.7.0 -d "$kfp_directory"
-
-# Download boto3 package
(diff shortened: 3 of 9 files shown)


tests/resources/test-pipeline-with-custom-pip-server.py
```diff
@@ -3,10 +3,12 @@
 # Edited the compiled version manually, to remove the --trusted-host flag
 # this is so we can test for tls certs validations when launcher installs packages
 @dsl.component(base_image="quay.io/opendatahub/ds-pipelines-ci-executor-image:v1.0",
-packages_to_install=['boto3'],
-pip_index_urls=['https://nginx-service.test-pypiserver.svc.cluster.local:443/simple/'])
+packages_to_install=['numpy'],
+pip_index_urls=['https://nginx-service.test-pypiserver.svc.cluster.local/simple/'],
+pip_trusted_hosts=['nginx-service.test-pypiserver.svc.cluster.local'])
```
HumairAK, line 8: we don't want to mark this as trusted, we want to do a tls verification, this should be empty as per your change [here](https://github.com/kubeflow/pipelines/pull/11151#discussion_r1738959914)

.github/scripts/python_package_upload/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM docker.io/python:3.9
```
HumairAK, line 1: what is the purpose of this dockerfile?

tests/resources/test-pipeline-with-custom-pip-server.py
diegolovison: This is out of the scope of this PR.
Created a new JIRA https://issues.redhat.com/browse/RHOAIENG-13980

.github/scripts/python_package_upload/Dockerfile
diegolovison: Updated the PR description

tests/resources/test-pipeline-with-custom-pip-server.py
HumairAK: how is this out of scope? the PR is explicitly removing tls verification

.github/scripts/python_package_upload/Dockerfile
HumairAK: that doesn't really answer the question, where is this dockerfile being used? 
are we building this anywhere? maintaining this image anywhere? 

docs on how one should use this? 

right now it's just a random dockerfile sitting in the repo

diegolovison: > where is this dockerfile being used? are we building this anywhere? 

Look at `.github/scripts/python_package_upload/package_upload_run.sh` 

> maintaining this image anywhere?

No. Just for testing purposes.

> docs on how one should use this?

By looking at `package_upload_run.sh` do still believe I should document? If yes, I can do it.

> right now it's just a random dockerfile sitting in the repo

Can you elaborate more what exactly it means?

HumairAK: I see, I missed that thanks. Do we have to build this every action? My concern is that we have [limited space](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) in the default runners, can we benefit from buildling this and uploading it to a quay repo instead?

tests/resources/test-pipeline-with-custom-pip-server.py
diegolovison: Rephrased the title and the description to explicit install https://issues.redhat.com/browse/RHOAIENG-13980

.github/scripts/python_package_upload/Dockerfile
diegolovison: If we get impacted by the limited space we can find out an alternative. 

By uploading to a quay repo we need to upload a new image version for every KFP release. The current way do it in runtime.

HumairAK: works for me!

HumairAK:
openshift-ci test failures look unrelated, ignoring this for now

PR #725, opened 2024-10-18
fix: update oauth-proxy-image and MLMD Envoy images

jgarciao:
Setting the same image that rhoai dashboard is currently using

## The issue resolved by this Pull Request:
Resolves [RHOAIENG-14652](https://issues.redhat.com/browse/RHOAIENG-14652) and [RHOAIENG-11441](https://issues.redhat.com/browse/RHOAIENG-11441)

## Description of your changes:
Updates the oauth-proxy image to use version [v4.14.0-202408260910.p0.ga4a2f27.assembly.stream.el8](https://catalog.redhat.com/software/containers/openshift4/ose-oauth-proxy/5cdb2133bed8bd5717d5ae64?image=66cefc14401df6ff4664ec43&container-tabs=gti), in order to fix a compatibility issue with OpenShift 4.16+ FIPS clusters

Note: this PR is setting the same image that the dashboard team used to the same error recently in [RHOAIENG-10450](https://issues.redhat.com/browse/RHOAIENG-10450), but it is possible that it will have to be updated again once the initiative to standardize these images is developed (more info at [RHOAIENG-11357](https://issues.redhat.com/browse/RHOAIENG-11357))

Updates also the MLMD Envoy image. New version is [2.5.5-6](https://catalog.redhat.com/software/containers/openshift-service-mesh/proxyv2-rhel8/5d2cda455a134672890f640a?image=66f66f4e5f86ba811464415f)

## Testing instructions
Follow the testing instructions at [RHOAIENG-14652](https://issues.redhat.com/browse/RHOAIENG-14652). Contact @jgarciao if you need a FIPS cluster

## Checklist
- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work
   - I could not build dspo locally. I tested it manually modifying the image in the dspa Deployment

PR #741, opened 2024-11-08
update ouath/envoy img to match params in generator

HumairAK:


HumairAK:
skipping ci as this change doesn't impace any images used in ci (don't use oauth proxy, and use upstream envoy instead of the one from rh registry)

jgarciao:
lgtm (the images are the same than the ones set recently in https://github.com/opendatahub-io/data-science-pipelines-operator/pull/725)

PR #683, opened 2024-08-09
Added support for TLS to MLMD GRPC Server

hbelmiro:
## The issue resolved by this Pull Request:
Resolves https://issues.redhat.com/browse/RHOAIENG-4971

This PR depends on:

- https://github.com/opendatahub-io/data-science-pipelines/pull/72

## Description of the changes:
Added support for TLS to MLMD GRPC Server

## Testing instructions

There are 2 scenarios for testing:

<details>
  <summary>TLS Enabled</summary>

1. Deploy the following DSPA

    ```yaml
    apiVersion: datasciencepipelinesapplications.opendatahub.io/v1alpha1
    kind: DataSciencePipelinesApplication
    metadata:
      name: dspa
    spec:
      dspVersion: v2
      podToPodTLS: true
      apiServer:
        image: "quay.io/opendatahub/ds-pipelines-api-server:pr-72"
        argoDriverImage: "quay.io/opendatahub/ds-pipelines-driver:pr-72"
        argoLauncherImage: "quay.io/opendatahub/ds-pipelines-launcher:pr-72"
        enableSamplePipeline: true
      persistenceAgent:
        image: "quay.io/opendatahub/ds-pipelines-persistenceagent:pr-72"
      scheduledWorkflow:
        image: "quay.io/opendatahub/ds-pipelines-scheduledworkflow:pr-72"
      mlmd:  
        deploy: true  # Optional component
        grpc:
          image: "quay.io/opendatahub/mlmd-grpc-server:latest"
        envoy:
          image: "registry.redhat.io/openshift-service-mesh/proxyv2-rhel8:2.3.9-2"
      mlpipelineUI:
        deploy: true  # Optional component 
        image: "quay.io/opendatahub/ds-pipelines-frontend:pr-72"
      objectStorage:
        minio:
          deploy: true
          image: 'quay.io/opendatahub/minio:RELEASE.2019-08-14T20-37-41Z-license-compliance'
    ```
2. Run the sample pipeline
3. The run must complete successfully
</details>

<details>
  <summary>TLS Disabled</summary>

1. Deploy the following DSPA

    ```yaml
    apiVersion: datasciencepipelinesapplications.opendatahub.io/v1alpha1
    kind: DataSciencePipelinesApplication
    metadata:
      name: dspa
    spec:
      dspVersion: v2
      podToPodTLS: false
      apiServer:
        image: "quay.io/opendatahub/ds-pipelines-api-server:pr-72"
[...]

PR #765, opened 2024-12-30
WIP: Refactor reconcilor so each component is returning status

anishasthana:
Related issue: https://issues.redhat.com/browse/RHOAIENG-13321

I took a stab at doing what I _think_ the underlying issue is asking for. @hbelmiro let me know if I'm on the right track!

I haven't gone through and updated all the dependent functions yet since I don't want to go down the stack if my starting point is completely wrong.

Related: https://github.com/opendatahub-io/data-science-pipelines-operator/pull/683#discussion_r1769286038

### DIFF
--- a/controllers/dspipeline_controller.go
+++ b/controllers/dspipeline_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+
 	"github.com/opendatahub-io/data-science-pipelines-operator/controllers/dspastatus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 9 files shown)


dsp-developers:
  A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-765`

dsp-developers:
  A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-765`

controllers/dspipeline_controller.go
```diff
@@ -318,18 +319,18 @@ func (r *DSPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 
-		err = r.ReconcileWorkflowController(dspa, params)
+		_, err = r.ReconcileWorkflowController(dspa, params)
 		if err != nil {
 			return ctrl.Result{}, err
 		}
 
 		// MLMD should be the last to reconcile because it can cause an early exit due to the lack of the TLS secret, which may not have been created yet.
-		err = r.ReconcileMLMD(ctx, dspa, params)
+		_, err = r.ReconcileMLMD(ctx, dspa, params)
```
hbelmiro, line 333: At least here we should use the returned status to set it in `r.setStatus`.

I wonder if we shouldn't use (set it in `r.setStatus`) the returned status in all cases when `status != nil`. The difference is that the condition will end up with a more friendly message instead of the original error. WDYT?

anishasthana: As far as I can tell the setStatus functions all take actual metav1. conditions. How would we be using the returned status at all?

dsp-developers:
  A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-765`

controllers/dspipeline_controller.go
```diff
@@ -374,10 +375,11 @@ func (r *DSPAReconciler) setStatusAsUnsupported(conditionType string, err error,
 	setStatus(condition)
 }
 
-func (r *DSPAReconciler) setStatus(ctx context.Context, resourceName string, conditionType string,
+func (r *DSPAReconciler) setStatus(ctx context.Context, resourceName string, conditionType string, statusMessage string,
 	dspa *dspav1.DataSciencePipelinesApplication, setStatus func(metav1.Condition),
 	log logr.Logger) {
 	condition, err := r.evaluateCondition(ctx, dspa, resourceName, conditionType)
+	condition.Message = statusMessage
```
anishasthana, line 382: I'm not sure if overwriting the `Message` here is the right call. The reason I'm wondering that is that https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/controllers/dspipeline_controller.go#L406 has a number of instances in which it is setting the `Message` field itself. Thoughts @hbelmiro ?

hbelmiro: I actually proposed replacing
```go
r.setStatus(ctx, params.MlmdProxyDefaultResourceName, config.MLMDProxyReady, statusMessage, dspa,
				dspaStatus.SetMLMDProxyStatus, log)
```

with

```go
			condition, err := r.evaluateCondition(ctx, dspa, params.MlmdProxyDefaultResourceName, config.MLMDProxyReady)
			condition.Message = status
			if err != nil {
				r.setStatusAsNotReady(config.MLMDProxyReady, err, dspaStatus.SetMLMDProxyStatus)
				return ctrl.Result{}, err
			}
			dspaStatus.SetMLMDProxyStatus(condition)
```

[here](https://github.com/anishasthana/data-science-pipelines-operator/blob/54ea85ecd6100bd556e309ef1cfe1ad136960f4d/controllers/dspipeline_controller.go#L338).

PR #788, opened 2025-02-14
update params to match latest configs

HumairAK:
To keep it consistent with the latest config, we should really just reference the config in config/base/parsme.env instead to avoid having to do this. 

The shas are syncing with what's in main params
https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/config/base/params.env


hbelmiro (commented):
/lgtm

gmfrasca (commented):
/lgtm

PR #816, opened 2025-04-11
correct github octokit calls

HumairAK:
It seems whatever octokit rest versioning used in github-script 3.1.0 is the older one ([v17 at latest](https://octokit.github.io/rest.js/v17/)). It is not clear where the mismatch is within the[ github-script package.json ](https://github.com/actions/github-script/blob/releases/v3/package.json)

but this is consistent with how [we use it upstream](https://github.com/kubeflow/pipelines/blob/master/.github/workflows/add-ci-passed-label.yml#L32), and we don't have issues there 🤷🏾

### DIFF
--- a/.github/workflows/release_create.yaml
+++ b/.github/workflows/release_create.yaml
@@ -29,7 +29,7 @@ jobs:
         uses: actions/github-script@v3.1.0
         with:
           script: |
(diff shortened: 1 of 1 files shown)


mprahl (approved):
/approve

PR #837, opened 2025-05-19
Build the binary in FIPS mode

mprahl:
Resolves #<issue_number>

This also updates the MariaDB image in the examples to use one that supports TLS connections in FIPS mode. The previous one was using too old ciphers.

## Testing instructions

Deploy DSPO on a FIPS enabled cluster and run the Iris pipeline with a UBI 9 based image. For example, `registry.redhat.io/ubi9/python-311:9.6-1747333117`. This is required for the OpenSSL versions to match with the FIPS compiled launcher and argoexec.

## Checklist


Dockerfile
```diff
@@ -22,9 +22,9 @@ COPY controllers/ controllers/
 # the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
 USER root
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager main.go
```
gmfrasca, line 25: should we leave the GOOS param since it already has the same defaulted value of `linux`, and update the GOARCH similarly to use $TARGETARCH but default to `amd64`?  If not, should we remove the ARG layers since they're now unused?

mprahl: Good point. I didn't notice that this was different compared to the Konflux Dockerfiles. I guess I should have read the Code Rabbit review comment more closely lol.

tests/resources/dspa-lite.yaml
```diff
@@ -60,7 +60,7 @@ spec:
   database:
     mariaDB:
       deploy: true
-      image: quay.io/centos7/mariadb-103-centos7:ea07c0dade9571d78a272b453fd2dea92077dc7f
+      image: quay.io/sclorg/mariadb-105-c9s:latest
```
HumairAK, line 63: hrmm I also see: https://quay.io/repository/mariadb-foundation/mariadb-devel?tab=tags which looks like a more official source, do you know what the difference is?

mprahl: @HumairAK not sure but the SCL Org is a Red Hat project:
https://github.com/sclorg

HumairAK: works for me

gmfrasca (commented):
/lgtm

PR #847, opened 2025-05-27
replace standard flag with openshift-ci

HumairAK:
openshift-ci case is different than that of the kind case or rhoai case, it makes sense to just have this be it's own thing

for example, for openshift-ci we are not going to build the image and push to registry, openshift-ci has it's own functionality to support this so we can have the image be available prior to executing this test script


.github/scripts/tests/tests.sh
```diff
@@ -28,16 +28,21 @@ RESOURCES_DIR_CRD="${GIT_WORKSPACE}/.github/resources"
...
 get_dspo_image() {
-  if [ "$REGISTRY_ADDRESS" = "" ]; then
-    # this function is called by `IMG=$(get_dspo_image)` that captures the standard output of get_dspo_image
-    set -x
-    echo "REGISTRY_ADDRESS variable not defined."
-    exit 1
+  if [ ! -z "$DSPO_IMAGE_REF" ]; then
+    echo $DSPO_IMAGE_REF
+  else
+    if [ "$REGISTRY_ADDRESS" = "" ]; then
```
gmfrasca, line 37:
```suggestion
    if [ -z "$REGISTRY_ADDRESS" ]; then
```

just for consistency

gmfrasca (commented):
/lgtm

PR #848, opened 2025-05-29
chore: parameterize container cli for test.sh

HumairAK:
this is so we can specify podman in the openshift-ci environment


hbelmiro (commented):
/lgtm

PR #851, opened 2025-05-30
provide option to upload packages directly

HumairAK:
instead of requiring a container, in openshift ci running podman is severely limited, and docker is not available

I'm also not sure what the initial intention was for the `--python` flag, but this seems pretty useless, I think the original author intended to use `--python-version`, i.e. download the package for a specific python version, `--python` just specifies which python interpreter to run pip with

I was encountering issues with running this command in my local environment as well: 

```
ERROR: The --python option must be placed before the pip subcommand name
```

## Summary by CodeRabbit

- **Chores**
  - Simplified package download process by removing Python version-specific downloads.
  - Updated environment variable for controlling container usage during package upload and testing, replacing image specification with a boolean flag.


anishasthana (approved):
/lgtm

PR #1000, opened 2026-03-26
Add Plugins Support

alyssacgoins:
Resolves #<issue_number>

Add support for plugins, following https://github.com/kubeflow/pipelines/pull/13005 and https://github.com/kubeflow/pipelines/pull/13052

## Checklist

## Summary by CodeRabbit

* **New Features**
  * MLflow integration added for Data Science Pipelines Applications.
  * CRD exposes spec.apiServer.mlflow with IntegrationMode (AUTODETECT, DISABLED) and InjectUserEnvVars flag.
  * API server can auto-detect and inject MLflow endpoint into its runtime config when enabled.
  * RoleBindings added to grant MLflow integration permissions to the API server and pipeline runner.
  * Runtime now watches MLflow resources to enable integration.

* **Chores**
  * Release parameters updated to include an MLFLOW option.


api/v1/dspipeline_types.go
```diff
@@ -55,6 +55,13 @@ type DSPASpec struct {
 	// Proxy configuration for all DSPA components to enable usage in environments requiring proxy access
 	// +kubebuilder:validation:Optional
 	Proxy *ProxyConfig `json:"proxy,omitempty"`
+
+	// Plugins is a string list of JSON configs for each API server plugin
+	Plugins string `json:"plugins,omitempty"`
+
+	// Enable plugins. Defaults to True.
+	// +kubebuilder:validation:Optional
+	PluginsEnabled bool `json:"pluginsEnabled,omitempty"`
```
hbelmiro: What's the reason for having `PluginsEnabled`?
It's not a strong opinion, but I'd leave only `Plugins`. If `Plugins` is empty or omitted, plugins are disabled.
I can see having `PluginsEnabled` useful when we have several plugins and we want to disable plugins temporarily without having to set them all again later, which is not the case for now.
If we happen to have several plugins in the future, we can add the filter later.

hbelmiro (commented):
- Can we move the MLflow check to the reconciler? That way CRs can be added/modified on the fly and users can have status feedback.
- I'd remove `PluginsEnabled`. If `Plugins` is empty or absent, plugins are disabled.
- Probably out of scope: It would be nice to have admin (operator level) granularity on what plugins are allowed by DSPAs

controllers/dspipeline_controller.go
```diff
@@ -304,6 +304,27 @@ func (r *DSPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		dspaStatus.SetObjStoreReady()
 	}
 
+	if params.PluginsEnabled && params.Plugins != "" {
```
HumairAK: I think your approach needs to change here. Take a look at how plugin settings are configured at the [API server level](https://github.com/kubeflow/pipelines/blob/master/proposals/12862-mlflow-integration/README.md#mlflow-configuration). This is the config that is present [here](https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/config/internal/apiserver/default/server-config.yaml.tmpl). This is what you need to update when mlflow integrationMode is set to AutoDetect and you the MLflow instance exists on the cluster (i.e. you have successfully retrieved the mlflow endpoint). 

Now the only thing you need to dynamically fetch is the endpoint (as mentioned earlier) from the MLflow CR. The rest you can infer/hardcode (as constants): 

Timeout: 30s (or have llm suggest a reasonable timeout) 
TLS: the dspa custom bundle path is sufficient
WorkspacesEnabled: true
AuthType: kubernetes
ExperimentDescription: "Created by AI Pipelines."

This can go straight in to the pipeline server config, and the integration will be enabled.

HumairAK (approved):

HumairAK (changes requested):

HumairAK (changes requested):

api/v1/dspipeline_types.go
```diff
@@ -466,6 +468,24 @@ type ComponentDetailStatus struct {
...
+const (
+	AutoDetect IntegrationMode = "AUTODETECT"
+	Disabled   IntegrationMode = "DISABLED"
+)
+
+type MLflowConfig struct {
+	// Indicates whether to enable MLflow plugin on API server if plugin installed on cluster,
+	// or disable entirely.
+	// +kubebuilder:validation:Optional
+	IntegrationMode IntegrationMode `json:"integrationMode,omitempty"`
```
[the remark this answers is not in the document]
alyssacgoins: Updated, added // +kubebuilder:default=false for `InjectUserEnvVars` as well.

config/internal/apiserver/default/server-config.yaml.tmpl
```diff
@@ -25,4 +25,6 @@ data:
       "ARCHIVE_CONFIG_LOG_PATH_PREFIX": "/artifacts",
       "InitConnectionTimeout": "6m"{{ if .APIServerWorkspaceJSON }},
       "Workspace": {{ .APIServerWorkspaceJSON }}{{ end }}
+      "Plugins": {
+        "mlflow": {{ if .APIServerPluginsJson }}
```
HumairAK: is this resulting in a valid json? it looks like it's missing a close `}`

Can we add a unit test for the rendering of server-config.yaml.tmpl with the new MLflow config? The declarative test framework already supports this, it should just be a matter of adding an expected ConfigMap to one of the existing test cases under controllers/testdata/declarative/, or adding a focused assertion in apiserver_test.go. Minimal effort given the existing infrastructure.

main.go
```diff
@@ -259,7 +274,18 @@ func main() {
 				},
 			},
 		},
-	})
+	}
+
+	if isCRDAvailable(restCfg, mlflowv1.GroupVersion.String()) {
+		mlflowSyncPeriod := 5 * time.Minute
+		mgrOpts.Cache.ByObject[&mlflowv1.MLflow{}] = cache.ByObject{SyncPeriod: &mlflowSyncPeriod}
+		setupLog.Info("MLflow CRD detected, caching MLflow resources", "syncPeriod", mlflowSyncPeriod)
```
HumairAK: (AI Curated response)

The isCRDAvailable check in main.go only runs at operator startup. If the MLflow CRD is installed after DSPO is already running, it won't be detected and the cache won't be configured — requiring an operator restart.                                  
                                                                                                                                                                                                                                                            
  Instead of relying on the manager cache for the MLflow CR, consider fetching the MLflow endpoint directly via an uncached API call in RetrieveMLflowEndpoint and caching the result in-memory with a TTL (e.g. a simple struct with sync.Mutex, the cached
   value, and an expiry timestamp). This way:                                                                                                                                                                                                               
                  
  1. No startup-time CRD detection needed — the lookup happens lazily during reconciliation.                                                                                                                                                                
  2. The MLflow CRD/CR can be installed at any time without restarting DSPO.
  3. The TTL controls how often you hit the API server, giving you the same performance benefit as the cached approach.                                                                                                                                     
                                                                                                                                                                                                                                                            
  The mlflowv1 scheme registration can stay since it's needed to deserialize the MLflow CR, but the manager cache configuration and isCRDAvailable check in main.go could be removed.

HumairAK (changes requested):

controllers/dspipeline_controller.go
```diff
@@ -201,6 +201,8 @@ func (r *DSPAReconciler) DeleteResourceIfItExists(ctx context.Context, obj clien
 //+kubebuilder:rbac:groups=workload.codeflare.dev,resources=appwrappers;appwrappers/finalizers;appwrappers/status,verbs=create;delete;deletecollection;get;list;patch;update;watch
 //+kubebuilder:rbac:groups=pipelines.kubeflow.org,resources=pipelines;pipelines/finalizers,verbs=create;get;list;watch;update;patch;delete
 //+kubebuilder:rbac:groups=pipelines.kubeflow.org,resources=pipelineversions;pipelineversions/status;pipelineversions/finalizers,verbs=create;get;list;watch;update;patch;delete
+//+kubebuilder:rbac:groups=mlflow.kubeflow.org,resources=experiments,verbs=create;get;list
```
VaniHaripriya, line 212: Controller might not require access to MLflow experiments/runs, access to MLflow CRs should suffice.

controllers/dspipeline_params.go
```diff
@@ -729,6 +789,24 @@ func (p *DSPAParams) ExtractParams(ctx context.Context, dsp *dspa.DataSciencePip
...
 
+		if p.MLflow != nil && *p.MLflow.IntegrationMode == dspa.AutoDetect {
+			// Retrieve internal MLflow service endpoint for use in API Server MLflow plugin config.
+			mlflowEndpoint, err := p.RetrieveMLflowEndpoint(ctx, client, log)
+			if err == nil {
+				apiServerExternalUrl, err := util.GetRouteHostname(ctx, p.APIServerServiceName, p.Namespace, client)
+				if err != nil {
+					log.Info("Unable to retrieve API server route for KFP base URL", "error", err)
+				}
+				pluginCfg, err := BuildMLflowPluginConfigJson(mlflowEndpoint, p.CustomCABundleRootMountPath, *p.MLflow.InjectUserEnvVars, apiServerExternalUrl)
```
VaniHaripriya, line 823: MLflow plugin JSON is built before [APIServer custom CA mount overrides](https://github.com/opendatahub-io/data-science-pipelines-operator/pull/1000/changes#diff-5fd0ead87d572dec71ec0194647f49e295c220aa3798b993d75e3b491f64a17aL732-L734) are applied. If spec.apiServer.caBundleFileMountPath is set, plugin config may still contain the default mount path.

controllers/dspipeline_params.go
```diff
@@ -703,6 +747,22 @@ func (p *DSPAParams) ExtractParams(ctx context.Context, dsp *dspa.DataSciencePip
 		p.PodToPodTLS = *dsp.Spec.PodToPodTLS
 	}
 
+	if dsp.Spec.MLflow != nil {
+		defaultMode := dspa.AutoDetect
+		mlflowCfg := dspa.MLflowConfig{
+			IntegrationMode:   &defaultMode,
+			InjectUserEnvVars: dsp.Spec.MLflow.InjectUserEnvVars,
```
VaniHaripriya, line 762: Could you add nil check for `spec.mlflow.integrationMode `and `spec.mlflow.injectUserEnvVars`

VaniHaripriya (changes requested):

dsp-developers:
  A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-1000`

PR #1048, opened 2026-05-14
Add mlflow plugin support

VaniHaripriya:
Resolves [RHOAIENG-53614](https://redhat.atlassian.net/browse/RHOAIENG-53614)

## Checklist

## Summary by CodeRabbit

* **New Features**
  * MLflow integration added to the Data Science Pipelines spec with selectable mode (AUTODETECT/DISABLED) and optional injectUserEnvVars (default: AUTODETECT/false).
  * API server can generate MLflow plugin config (conditional) and controller discovers MLflow endpoints with caching; workload RBAC is enabled when plugin config is emitted.
  * Sample manifest demonstrating MLflow-enabled deployment added.

* **Tests**
  * Unit tests for MLflow plugin JSON generation, endpoint discovery/validation, and API server readiness checks.


dsp-developers:
A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-1048`

```bash
cd $(mktemp -d)
git fetch origin pull/1048/head
git checkout -b pullrequest 364891a8a2eb75e5dfd07ebca0e7ec4ddb14010c
oc new-project opendatahub
make deploy IMG="quay.io/opendatahub/data-science-pipelines-operator:pr-1048"
```

dsp-developers:
  A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:pr-1048`

controllers/dspipeline_params.go
```diff
@@ -225,6 +254,42 @@ func (p *DSPAParams) RetrieveSecret(ctx context.Context, client client.Client, s
...
+	if err != nil {
+		log.V(1).Info(fmt.Sprintf("Unable to retrieve mlflow resource [%s].", mlflowName))
+		return "", err
+	}
+	status := mlflow.Status
+	if status.Address != nil && status.Address.URL != "" {
+		return mlflow.Status.Address.URL, nil
+	}
+	return "", errors.New("MLflow resource missing Status.Address.URL field. Unable to resolve endpoint")
+}
```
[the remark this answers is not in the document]
VaniHaripriya: Cleaned up temporary workaround added for testing.

controllers/dspipeline_controller.go
```diff
@@ -207,6 +207,9 @@ func (r *DSPAReconciler) DeleteResourceIfItExists(ctx context.Context, obj clien
 //+kubebuilder:rbac:groups=workload.codeflare.dev,resources=appwrappers;appwrappers/finalizers;appwrappers/status,verbs=create;delete;deletecollection;get;list;patch;update;watch
 //+kubebuilder:rbac:groups=pipelines.kubeflow.org,resources=pipelines;pipelines/finalizers,verbs=create;get;list;watch;update;patch;delete
 //+kubebuilder:rbac:groups=pipelines.kubeflow.org,resources=pipelineversions;pipelineversions/status;pipelineversions/finalizers,verbs=create;get;list;watch;update;patch;delete
+//+kubebuilder:rbac:groups=mlflow.opendatahub.io,resources=mlflows,verbs=get;list;watch
+//+kubebuilder:rbac:groups=mlflow.kubeflow.org,resources=experiments,verbs=create;get;list;update
+//+kubebuilder:rbac:groups=mlflow.kubeflow.org,resources=runs,verbs=create;get;update;list
```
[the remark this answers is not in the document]
VaniHaripriya: The API server requires update permissions to trigger MLflow APIs.

api/v1/dspipeline_types.go
```diff
@@ -55,6 +55,10 @@ type DSPASpec struct {
 	// Proxy configuration for all DSPA components to enable usage in environments requiring proxy access
 	// +kubebuilder:validation:Optional
 	Proxy *ProxyConfig `json:"proxy,omitempty"`
+
+	// MLflow configuration for MLflow API Server plugin
+	// +kubebuilder:validation:Optional
+	MLflow *MLflowConfig `json:"mlflow,omitempty"`
```
hbelmiro, line 63: MLflow integration is opt-in, but the acceptance criteria require opt-out. The `mlflow` field is an optional pointer (`*MLflowConfig`) with `omitempty`. When omitted from the DSPA spec, `p.MLflow` is nil and the entire MLflow block at `dspipeline_params.go:844` is skipped (`if p.MLflow != nil && mlflowMode == dspa.AutoDetect`). The AC says "DSPA CRD needs a new field to **opt out** of the MLflow integration" — meaning auto-detect should be the default behavior and the field should only be needed to disable it. Currently, users must explicitly add `mlflow: {}` to their DSPA spec to enable any detection at all.

config/internal/apiserver/default/role_ds-pipeline.yaml.tmpl
```diff
@@ -103,3 +109,21 @@ rules:
       - get
     resources:
       - routes
+  - apiGroups:
```
hbelmiro, line 113: **RHOAIENG-53614 acceptance criteria gap: MLflow RBAC rules are unconditional.** The `mlflow.kubeflow.org` permissions for `experiments` and `runs` are hardcoded in the role templates and applied to every DSPA, regardless of whether MLflow is configured or installed. The same applies to `role_pipeline-runner.yaml.tmpl:145-161`. This violates least-privilege — every API server and pipeline-runner service account gets create/get/list/update on MLflow resources even when there is no MLflow integration.

**Why it matters:** The AC says "pipeline-runner and API server service accounts need access to the MLflow instance" — implying this access should be scoped to when the integration is active. Additionally, AC 5 says "handle the case where MLflow is uninstalled to automatically remove the integration." The unconditional RBAC means uninstalling MLflow leaves these permission rules in place on every DSPA's roles.

**Suggested fix:** Make the MLflow RBAC rules conditional in the templates using a Go template guard:

controllers/dspipeline_params.go
```diff
@@ -752,6 +836,56 @@ func (p *DSPAParams) ExtractParams(ctx context.Context, dsp *dspa.DataSciencePip
...
+			apiServerReady, readinessErr := p.IsAPIServerDeploymentReady(ctx, client)
+			if readinessErr != nil {
+				log.Error(readinessErr, "failed to determine APIServer readiness. MLflow API server plugin config generation deferred.")
+			} else if !apiServerReady {
+				log.V(1).Info("APIServer deployment is not ready yet. Deferring MLflow API server plugin config generation.")
+			} else {
+				// Retrieve internal MLflow service endpoint for use in API Server MLflow plugin config.
+				mlflowEndpoint, err := p.RetrieveMLflowEndpoint(ctx, client, log)
+				if err == nil {
+					apiServerExternalURL, routeErr := util.GetRouteHostname(ctx, p.APIServerServiceName, p.Namespace, client)
```
hbelmiro: Redundant and misleading double-log when route retrieval fails. When `GetRouteHostname` returns a non-nil error, `apiServerExternalURL` is `""`. The code logs the error at line 855, then falls through to the `apiServerExternalURL == ""` check at line 858, logging "APIServer route is not available yet. Deferring..." — which mischaracterizes an actual error as a timing issue.

controllers/dspipeline_params.go
```diff
@@ -149,6 +154,30 @@ type ObjectStorageConnection struct {
...
 
+type PluginConfig struct {
+	Endpoint string               `json:"endpoint,omitempty" mapstructure:"endpoint"`
+	Timeout  string               `json:"timeout,omitempty" mapstructure:"timeout"`
+	TLS      TLSConfig            `json:"tls,omitempty" mapstructure:"tls"`
+	Settings MLflowPluginSettings `json:"settings,omitempty" mapstructure:"settings"`
+}
+
+type TLSConfig struct {
+	InsecureSkipVerify bool   `json:"insecureSkipVerify,omitempty" mapstructure:"insecureSkipVerify"`
```
hbelmiro: `InsecureSkipVerify` uses `omitempty` on a security-sensitive boolean. When set to `false` (the secure default, which is what `BuildMLflowPluginConfigJson` always produces), the field is omitted from the serialized JSON. If the consuming API server plugin does not default absent booleans to `false`, TLS verification could be silently skipped. Similarly, `InjectUserEnvVars` at line 178 has the same `omitempty` tag — when `false`, it will be absent from the JSON output.

**Suggested fix:** Remove `omitempty` from boolean fields that have meaningful zero values, especially security-relevant ones.

controllers/dspipeline_params.go
```diff
@@ -1000,3 +1134,31 @@ func (p *DSPAParams) ExtractParams(ctx context.Context, dsp *dspa.DataSciencePip
 
 	return nil
 }
+
+func BuildMLflowPluginConfigJson(mlflowEndpoint string, caBundlePath string, injectUserEnvVars bool, kfpBaseURL string) (string, error) {
```
hbelmiro, line 1128: No unit tests for any of the three new public functions: `BuildMLflowPluginConfigJson`, `RetrieveMLflowEndpoint`, and `IsAPIServerDeploymentReady`. `BuildMLflowPluginConfigJson` is a pure function with no external dependencies — it is straightforward to unit test and should cover at minimum: the default case, `injectUserEnvVars=true`, and the shape of the output JSON. `RetrieveMLflowEndpoint` and `IsAPIServerDeploymentReady` use `client.Client` and can be tested with a fake client as done elsewhere in this codebase (e.g., `apiserver_test.go`).

**Why it matters:** The MLflow plugin config is rendered into the API server's `server-config.json` ConfigMap. A malformed or missing JSON field can silently break the MLflow integration at runtime with no signal until a user tries to use it. Tests would catch regressions in JSON structure, field presence, and zero-value handling (see Comment 4).

main.go
```diff
@@ -264,7 +279,18 @@ func main() {
 				},
 			},
 		},
-	})
+	}
+
+	if isCRDAvailable(restCfg, mlflowv1.GroupVersion.String()) {
```
hbelmiro: **Acceptance criteria gap: MLflow CRD detection is startup-only.** `isCRDAvailable()` runs once during operator startup at `main.go:284`. If the MLflow CRD is installed after the operator is already running, the cache configuration won't be updated — no informer will be created for MLflow resources. The operator will still function via direct API calls (`DisableFor` path), but it won't receive watch events for MLflow CR changes and depends entirely on DSPA reconcile triggers to notice MLflow availability.

**Why it matters:** In a typical install flow, MLflow may be installed after DSPO. The operator will work (direct API calls succeed), but performance and responsiveness degrade compared to the cached path. Also, if the MLflow CRD is uninstalled while the operator is running, the existing cache informer will log repeated errors until the operator is restarted.

scripts/release/params.py
```diff
@@ -50,7 +50,8 @@
     "MANAGEDPIPELINES": "\"{}\"",
     "PLATFORMVERSION": "\"v0.0.0\"",
     "FIPSENABLED": "false",
-    "WEBHOOK_ANNOTATIONS": ""
+    "WEBHOOK_ANNOTATIONS": "",
+    "MLFLOW": "\"{}\"",
```
hbelmiro: The `MLFLOW` env var entry appears unused. The MLflow configuration is sourced from the DSPA CR spec (`dsp.Spec.MLflow`), not from operator-level environment variables or viper config. There is no corresponding `viper.Get("MLFLOW")` or `os.Getenv("MLFLOW")` call in the Go code. If this entry is not consumed, it should be removed to avoid confusion.

dsp-developers:
  A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:odh-pr-1048`

api/v1/dspipeline_types.go
VaniHaripriya: Good point, fixed this. No need for mlflow: {} to enable detection, omitting spec.mlflow now defaults to AUTODETECT .

controllers/dspipeline_params.go
VaniHaripriya: Updated this, so a route lookup error is logged once and we no longer fall through to the “not available yet” defer message.

controllers/dspipeline_params.go
VaniHaripriya: Removed omitempty from `InsecureSkipVerify` and `InjectUserEnvVars` so false is always emitted in the plugin JSON instead of being omitted.

controllers/dspipeline_params.go
VaniHaripriya: Added unit tests.

scripts/release/params.py
VaniHaripriya: Removed the `MLFLOW` env var.

config/internal/apiserver/default/role_ds-pipeline.yaml.tmpl
VaniHaripriya: Added a conditional Go template guard to the MLflow RBAC rules in the templates to control when they are applied.

main.go
VaniHaripriya: Removed startup isCRDAvailable() ,  MLflow is always on Client.Cache.DisableFor, with endpoint lookups via APIReader and a reconciler TTL cache behavior no longer depends on CRD presence at operator start.

dsp-developers:
  A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:odh-pr-1048`

controllers/dspipeline_params.go
```diff
@@ -225,6 +259,42 @@ func (p *DSPAParams) RetrieveSecret(ctx context.Context, client client.Client, s
 	return base64.StdEncoding.EncodeToString(secret.Data[secretKey]), nil
 }
 
+func (p *DSPAParams) RetrieveMLflowEndpoint(ctx context.Context, client client.Client, log logr.Logger) (string, error) {
```
hbelmiro: `RetrieveMLflowEndpoint` and `retrieveMLflowEndpointCached` (in `dspipeline_controller.go:90`) both perform the same MLflow endpoint lookup with near-identical logic but diverge subtly: the controller method uses `APIReader` and caches, while the params method uses the standard client and does not cache. Both hardcode the name `"mlflow"`. The `ExtractParams` call site at line 854 dispatches between them via a nil check on `ResolveMLflowEndpoint`.

The params method (`RetrieveMLflowEndpoint`) exists as a fallback when `ResolveMLflowEndpoint` is nil, but in production `main.go` always injects `retrieveMLflowEndpointCached`. The fallback path is only exercised in tests (and tests that test `RetrieveMLflowEndpoint` directly). This duplication means a bug fix in one method might not be applied to the other.

**Suggested fix:** Consider removing the fallback `RetrieveMLflowEndpoint` method and making `ResolveMLflowEndpoint` required. Tests can inject a stub directly.

controllers/dspipeline_controller.go
```diff
@@ -67,6 +72,61 @@ type DSPAReconciler struct {
...
+	}
+
+	r.mlflowEndpointCacheMu.Lock()
+	if r.mlflowEndpointCache == nil {
+		r.mlflowEndpointCache = map[string]mlflowEndpointCacheEntry{}
+	}
+	r.mlflowEndpointCache[namespace] = mlflowEndpointCacheEntry{
+		endpoint: mlflowObj.Status.Address.URL,
+		expires:  now.Add(ttl),
+	}
```
hbelmiro, line 155: Optional: The TTL cache never evicts stale entries. If the operator manages many namespaces over its lifetime, the `mlflowEndpointCache` map grows unbounded with entries from namespaces that have been deleted. Each entry is small (string + time), so this is not critical, but for long-running operators it is a slow leak.

controllers/dspipeline_params.go
```diff
@@ -752,6 +837,61 @@ func (p *DSPAParams) ExtractParams(ctx context.Context, dsp *dspa.DataSciencePip
...
+						}
+						effectiveCABundleFilePath := fmt.Sprintf("%s/%s", effectiveCABundleRootMountPath, effectiveCABundleFileName)
+						injectUserEnvVars := false
+						if p.MLflow.InjectUserEnvVars != nil {
+							injectUserEnvVars = *p.MLflow.InjectUserEnvVars
+						}
+						pluginCfg, err := BuildMLflowPluginConfigJson(mlflowEndpoint, effectiveCABundleFilePath, injectUserEnvVars, apiServerExternalURL)
+						if err != nil {
+							log.Info("Failed to build MLflow plugin config. MLflow API server plugin will not be enabled.")
+						}
```
hbelmiro: When `BuildMLflowPluginConfigJson` returns a non-nil error at line 883, the error is logged but `pluginCfg` (which is `""` on error) is still assigned to `p.APIServerPluginsJson` at line 886. This means `APIServerPluginsJson` stays `""`, `GrantMlflowWorkloadRBAC` stays `false`, and the plugin is correctly not enabled. So the behavior is correct by accident. But the log message says "MLflow API server plugin will not be enabled" while the code path falls through to setting `p.APIServerPluginsJson = pluginCfg` (which is empty). It would be clearer to `break`/`return` or add an `else` clause so the assignment only happens on success.

config/internal/apiserver/default/role_ds-pipeline.yaml.tmpl
```diff
@@ -50,6 +50,12 @@ rules:
     verbs:
       - get
       - list
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+    verbs:
+      - get
```
hbelmiro: The new unconditional `configmaps` get permission (lines 53-58) is added outside the `{{- if .GrantMlflowWorkloadRBAC }}` guard, meaning every DSPA gets this rule regardless of MLflow. If this configmap read is only needed for MLflow plugin config retrieval, it should be inside the guard.

dsp-developers:
  A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:odh-pr-1048`

dsp-developers:
  A new image has been built to help with testing out this PR: `quay.io/opendatahub/data-science-pipelines-operator:odh-pr-1048`

controllers/dspipeline_params.go
```diff
@@ -752,6 +815,62 @@ func (p *DSPAParams) ExtractParams(ctx context.Context, dsp *dspa.DataSciencePip
...
+					mlflowEndpoint string
+					err            error
+				)
+				if p.ResolveMLflowEndpoint == nil {
+					log.Info("ResolveMLflowEndpoint is not configured. Deferring MLflow API server plugin config generation.")
+				} else {
+					mlflowEndpoint, err = p.ResolveMLflowEndpoint(ctx, p.Namespace, log)
+					if err != nil {
+						log.Error(err, "failed to retrieve MLflow internal endpoint. MLflow API server plugin will not be enabled.")
+					} else {
```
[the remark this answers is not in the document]
VaniHaripriya: Added mlflow endpoint validation.

config/internal/apiserver/default/role_ds-pipeline.yaml.tmpl
VaniHaripriya: Updated the permissions with gaurd.

controllers/dspipeline_params.go
VaniHaripriya: Addressed this.

controllers/dspipeline_params.go
VaniHaripriya: Addressed this.

hbelmiro (approved):
/lgtm
/approve

PR #1071, opened 2026-07-03
fix: Use DeepCopy in cache transform functions to prevent data races

ugiordan:
## Summary

Fixes a data race in `stripManagedFields`, `stripConfigMapData`, and `stripSecretData` transform functions introduced in #992.

The `cache.TransformFunc` contract requires that transform functions must not modify the input object in place. The SharedIndexInformer keeps references to both old and new objects during Update events. Mutating in-place corrupts the old-object reference, causing data races under concurrent reconciliation.

**Fix**: Return a `DeepCopy()` with modifications instead of mutating the original pointer.

## Changes

- `stripManagedFields`: DeepCopyObject before mutating, return copy
- `stripConfigMapData`: DeepCopy before mutating, return copy
- `stripSecretData`: DeepCopy before mutating, return copy

## Ref

- Reported by @mwaykole in https://github.com/opendatahub-io/odh-model-controller/pull/757#discussion_r3519838706
- Same fix applied to odh-model-controller and kubeflow

### DIFF
--- a/main.go
+++ b/main.go
@@ -65,7 +65,9 @@ var (
 // stripManagedFields removes managed fields from cached objects to reduce memory.
 func stripManagedFields(i interface{}) (interface{}, error) {
 	if obj, ok := i.(client.Object); ok {
-		obj.SetManagedFields(nil)
+		cp := obj.DeepCopyObject().(client.Object)
+		cp.SetManagedFields(nil)
+		return cp, nil
 	}
 	return i, nil
 }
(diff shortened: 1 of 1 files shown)


ugiordan:
Closing this. Since client-go v1.27 (v0.27+), TransformFunc sees the object before any other actor, and in-place mutation is explicitly safe. See https://github.com/kubernetes/client-go/blob/v0.35.2/tools/cache/delta_fifo.go#L148-L164. DeepCopy adds unnecessary allocation overhead. DSPO uses client-go v0.35.3 where this guarantee holds.

PR #1063, opened 2026-06-10
feat: integrate with cluster TLS security profile

ugiordan:
## The issue resolved by this Pull Request:
https://issues.redhat.com/browse/RHOAIENG-61066

## Description of your changes:

Integrate data-science-pipelines-operator with the cluster-wide TLS security profile from `apiservers.config.openshift.io/cluster`, required for OCP 5.0 TLS profile compliance ([OCPSTRAT-2611](https://redhat.atlassian.net/browse/OCPSTRAT-2611)).

Uses `controller-runtime-common/pkg/tls` (the standard OCP library for TLS profile integration):

1. Register `configv1` with the scheme
2. Fetch TLS profile at startup via `FetchAPIServerTLSProfile`
3. Build `tls.Config` callback via `NewTLSConfigFromProfile`
4. Apply to both webhook (9443) and metrics server `TLSOpts`
5. Set `NextProtos` (`h2`, `http/1.1`) on all TLS endpoints
6. Register `SecurityProfileWatcher` that triggers graceful restart on profile change
7. Add RBAC for `config.openshift.io/apiservers` get/list/watch

Reference PRs:
- [openshift/cluster-machine-approver#286](https://github.com/openshift/cluster-machine-approver/pull/286) (canonical controller-runtime-common example)
- [opendatahub-io/mlflow-operator#140](https://github.com/opendatahub-io/mlflow-operator/pull/140) (RHOAI template)

Integration guide: https://docs.google.com/document/d/1XUehWMzXeteHPwuEyh96m7ORvTanCHPyKgSEx630UQQ/edit

## Testing instructions

- `go build ./...` passes
- `go vet ./...` clean
- `go test --tags=test_unit ./...` passes (94s, all controllers tests pass)
- `make manifests` regenerates RBAC with new `apiservers` permission
- CodeRabbit review: 0 findings on changed files

## Checklist
- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

## Summary by CodeRabbit

* **Chores**
  * Updated dependencies to align with OpenShift and Kubernetes APIs.
  * Expanded RBAC permissions so the controller can read API server configuration needed for TLS compliance.

* **Bug Fixes**
  * Metrics and webhook servers now use the cluster’s TLS security profile (with a safe fallback when not available).
[...]

PR #1072, opened 2026-07-06
fix: handle transient API errors in TLS profile resolution

ugiordan:
Add IsForbidden, IsServiceUnavailable, IsTimeout, and IsTooManyRequests to the graceful fallback cases when reading the cluster TLS profile. Add 10s context timeout to prevent hanging on slow API server responses.

Previously transient errors caused a fatal startup error, turning minor network issues or restricted RBAC into CrashLoopBackOff.

Follow-up to #1063 (TLS profile integration).

[RHOAIENG-61066](https://redhat.atlassian.net/browse/RHOAIENG-61066)

## Summary by CodeRabbit

* **Bug Fixes**
  * Startup now uses a bounded timeout when fetching the TLS profile, preventing indefinite waits during initialization.
  * Improved TLS profile fallback for restricted access and transient API conditions (forbidden access, service unavailable, timeouts, and rate limiting), reducing startup failures when the API is unavailable or throttling requests.
  * If the initial Kubernetes client setup fails, startup now continues with hardened TLS defaults instead of exiting immediately.

[RHOAIENG-61066]: https://redhat.atlassian.net/browse/RHOAIENG-61066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ugiordan:
Reopening - was accidentally closed during force-push

PR #1074, opened 2026-07-07
fix: TLS profile resilience (transient errors, adherence policy, watcher self-healing)

ugiordan:
## The issue resolved by this Pull Request:
Resolves [RHOAIENG-49909](https://redhat.atlassian.net/browse/RHOAIENG-49909)

## Description of your changes:
Follow-up to #1063 (TLS profile integration). Three improvements to TLS profile resolution at operator startup:

1. **Transient error handling**: `IsServiceUnavailable`, `IsTimeout`, and `IsTooManyRequests` errors fall back to the Intermediate TLS profile instead of crashing the operator. Prevents startup failures during temporary API server unavailability (e.g., cluster upgrades).

2. **TLSAdherencePolicy support**: Fetches the OCP 5.0 TLS adherence policy from the APIServer CR and wires it into the `SecurityProfileWatcher`. When the adherence policy changes, the operator restarts gracefully to pick up the new configuration.

3. **Watcher self-healing**: On transient errors, the `SecurityProfileWatcher` still registers (both for profile and adherence changes). When the API recovers, the watcher detects the real profile differs from the Intermediate fallback and triggers a graceful restart. No admin intervention needed.

Supersedes #1072 and #1073 (closed due to branch history issues).

## Testing instructions
- Deploy the operator on an OpenShift cluster with TLS profiles configured
- Verify the operator starts successfully under normal conditions
- Verify the operator starts with Intermediate profile fallback when the API server is temporarily unavailable
- Verify the operator self-heals (restarts) when the API server becomes available again

## Checklist
- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

[RHOAIENG-49909]: https://redhat.atlassian.net/browse/RHOAIENG-49909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1077, opened 2026-07-08
fix: gracefully handle forbidden error in TLS profile fetch

hbelmiro:
## Summary

- Handle `IsForbidden` errors gracefully in `fetchTLSProfile` by falling back to the Intermediate TLS profile instead of crashing the operator on startup
- Add the same defense-in-depth handling for the TLS adherence policy fetch in `main.go`
- Add unit test covering the forbidden error scenario using `WithInterceptorFuncs`

## Context

After PR #1063 introduced TLS profile fetching from the OpenShift APIServer resource, the operator crashes on startup when the service account lacks RBAC to read `apiservers` in `config.openshift.io`:

```
failed to fetch TLS profile: failed to get APIServer "/cluster": apiservers.config.openshift.io "cluster" is forbidden
```

PR #1074 added the missing ClusterRole entry, but the code itself had no graceful fallback for "forbidden" errors — it fell to the `default` branch which returned a hard error, causing `os.Exit(1)`. This fix makes the operator resilient regardless of RBAC state.

## Test plan

- [x] New `TestFetchTLSProfile_Forbidden` test passes
- [x] All existing TLS profile tests pass (no regressions)
- [x] Full `make unittest` passes
- [x] `golangci-lint run` clean on changed files

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fallback behavior when TLS policy information can’t be retrieved, including access-denied and unavailable-resource cases.
  * The app now safely continues with the default Intermediate TLS profile when OpenShift policy data is not accessible.
  * Error handling for policy lookup is more consistent, reducing the chance of unexpected failures while loading TLS settings.

### DIFF
--- a/main.go
+++ b/main.go
@@ -246,7 +246,7 @@ func main() {
 		tlsAdherence, adherenceErr = tlspkg.FetchAPIServerTLSAdherencePolicy(bootstrapCtx, bootstrapClient)
 		if adherenceErr != nil {
 			switch {
-			case apierrors.IsNotFound(adherenceErr), apimeta.IsNoMatchError(adherenceErr):
+			case apierrors.IsNotFound(adherenceErr), apimeta.IsNoMatchError(adherenceErr), apierrors.IsForbidden(adherenceErr):
 				setupLog.Info("APIServer TLS adherence policy unavailable")
 			case apierrors.IsServiceUnavailable(adherenceErr),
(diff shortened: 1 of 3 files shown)


hbelmiro:
Adding IsForbidden as a graceful fallback would mask the RBAC misconfiguration instead of surfacing it. If the RBAC is wrong, the operator should crash so the issue is visible.

PR #1079, opened 2026-07-16
fix: handle IsServerTimeout in TLS profile resolution

ugiordan:
## Description

`apierrors.IsServerTimeout` (StatusReasonServerTimeout) is a distinct Kubernetes error from `apierrors.IsTimeout` (StatusReasonTimeout). The former fires when the API server itself couldn't complete the request in time, the latter when a gateway/proxy times out.

Without handling `IsServerTimeout`, a server-side API timeout during TLS profile fetch falls to the default error case and prevents startup instead of gracefully falling back to Intermediate defaults.

One-line fix, no behavior change for the happy path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when fetching TLS profiles during server timeouts.
  * Requests now use an intermediate TLS profile fallback instead of failing in this scenario.


hbelmiro (approved):
/lgtm
/approve

PR #1082, opened 2026-07-30
Fix MLflow AUTODETECT not detecting stale/unavailable MLflow when mlflow-operator is down

VaniHaripriya:

When MLflow was previously deployed and then scaled down (along with mlflow-operator itself), DSPO continued to treat MLflow as available and kept the MLflow plugin enabled in the API server config. This caused pipeline run creation to fail with connection errors to a dead MLflow endpoint.
Root cause: `lookupMLflowEndpoint` only checked that `MLflow.Status.Address.URL` was set and syntactically valid. It never checked the `Available` condition or whether the underlying Deployment was actually ready. If `mlflow-operator` itself is also down, its last-reported status (including `Available=True`) is frozen and can no longer be trusted.

## Fix
- Check the MLflow CR's `Available` condition before trusting its reported endpoint.
- Additionally cross-check the live backing `mlflow` Deployment's `ReadyReplicas` directly via the Kubernetes API, so we don't rely solely on `mlflow-operator`'s self-reported status (which goes stale if `mlflow-operator` itself is down).

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MLflow endpoint validation to confirm the service is marked available and its deployment is fully ready before use.
  * Prevented pipelines from using stale, missing, or incomplete MLflow deployments.
  * Added safeguards for scaled-to-zero deployments and deployments with insufficient ready replicas.
  * Ensured endpoint lookup targets the configured MLflow instance rather than unrelated instances.

### DIFF
--- a/controllers/dspipeline_controller.go
+++ b/controllers/dspipeline_controller.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/opendatahub-io/data-science-pipelines-operator/controllers/dspastatus"
 	mlflowv1 "github.com/opendatahub-io/mlflow-operator/api/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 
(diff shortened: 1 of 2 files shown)


mprahl (approved):

mprahl:
Closing since on second thought, we decided to just make KFP more resilient to MLflow outages so that KFP doesn't restart 2 times on every MLflow upgrade.

PR #1095, opened 2026-09-04
fix(ci): always fall through on clone fallback when fork/branch not found

hbelmiro:
ls-remote returns 128 when a fork repo doesn't exist (e.g. contributor has no data-science-pipelines fork), which was treated as a hard transport error. Simplify check_remote_ref to binary found/not-found and make both fork and upstream tiers fall through on any failure. Tier 3 (default branch) catches genuinely missing repos via set -euo.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved build reliability when retrieving DSP images by continuing to try available fallback sources when an alternate source cannot be accessed.
  - Builds now report an error only after all configured sources have been exhausted, rather than stopping prematurely on an intermediate source failure.


alyssacgoins (approved):
/lgtm

PR #1103, opened 2026-09-11
fix: document DSPO database TLS exception and warnings

ugiordan:
## Summary

- Preserve compatibility for database TLS modes `false`, `skip-verify`, and `preferred`.
- Log warnings when database TLS is disabled, unverified, opportunistic, or unrecognized.
- Add the approved TLS scanner exception files for the intentional compatibility path.
- Keep verified database TLS available through `tls=true` with system trust or `spec.apiServer.cABundle`.
- Add direct regression coverage, public API documentation, generated CRD output, and a database TLS guide.

## Compatibility and security

Existing DSPAs using `tls=skip-verify` or `tls=preferred` continue to work, but the operator logs a warning. `skip-verify` disables certificate verification, and `preferred` can fall back to an unencrypted connection. Users requiring verified TLS should migrate to `tls=true` and provide a trusted CA through `spec.apiServer.cABundle` when needed.

The scanner finding for this intentional compatibility path is handled through the approved security exception process:
https://github.com/opendatahub-io/security-config/tree/main#tls-workflows

## TLS profile scope

The informational `tls-config-lint` finding in `controllers/storage.go` is intentionally unchanged. Replacing MinIO TLS settings with Go defaults would not honor centralized TLS adherence for Old or custom profiles. Correct remediation requires explicit centralized profile propagation to outbound clients and is outside this database compatibility change.

## Validation

- `make unittest`
- `go vet ./...`
- `make manifests`
- Pre-commit checks
- Generated CRD manifests are current

## Tracking

- Jira: https://redhat.atlassian.net/browse/RHOAIENG-94348
- Source scanner workflow PR: https://github.com/opendatahub-io/data-science-pipelines-operator/pull/987


controllers/storage.go
```diff
@@ -128,6 +126,11 @@ func getTransportWithProxyAndCACert(log logr.Logger, pemCerts [][]byte, proxyCon
 	if err != nil {
 		return nil, fmt.Errorf("error creating default transport : %s", err)
 	}
+	if secure {
+		// Let Go negotiate its supported TLS versions instead of inheriting the
+		// fixed minimum version from MinIO's default transport.
+		transport.TLSClientConfig = new(tls.Config)
```
alyssacgoins: Could theoretically allow TLS 1.0/1.1 if both client and server support only older versions (however, this is unlikely to occur)

controllers/config/defaults.go
```diff
@@ -259,6 +259,9 @@ func ValidateDBExtraParams(raw string) (map[string]string, error) {
 			return nil, fmt.Errorf("customExtraParams contains disallowed key %q; allowed keys: %v",
 				key, allowedKeysList())
 		}
+		if key == "tls" && (strings.EqualFold(params[key], "skip-verify") || strings.EqualFold(params[key], "preferred")) {
+			return nil, fmt.Errorf("customExtraParams tls=%q disables certificate verification; use tls=true with a trusted CA bundle", params[key])
```
alyssacgoins: This  update will break any currently deployed DSPAs that utilize tls=skip-verify or tls=preferred. Something like the following error message will give users more clarity on the sudden failure
```suggestion
			return false, fmt.Errorf(
			"database TLS mode %q disables certificate verification and is rejected for security compliance.\n\n"+
				"Migration Options:\n"+
				"1. For databases with self-signed or internal CA certificates:\n"+
				"   - Set customExtraParams to {\"tls\":\"true\"}\n"+
				"   - Add CA certificate via spec.apiServer.cABundle\n\n"+
				"2. For databases accessible over trusted networks without TLS:\n"+
				"   - Set customExtraParams to {\"tls\":\"false\"}\n\n"+
				"For more details, see: https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/docs/database-tls.md",
			tls)
```

alyssacgoins (changes requested):

docs/database-tls.md
```diff
@@ -0,0 +1,41 @@
...
+    cABundle:
+      configMapName: database-ca
+      configMapKey: ca.crt
+```
+
+## Migration from verification bypass modes
+
+The `skip-verify` and `preferred` modes are rejected because they bypass
+certificate verification. Before upgrading, replace either mode with `true`
+and configure `spec.apiServer.cABundle` when the database CA is not already
```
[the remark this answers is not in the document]
alyssacgoins: @ugiordan can you take a look at this?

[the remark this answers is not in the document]
ugiordan: @coderabbitai I have added a regression test proving `{"tls":true}` is rejected while `{"tls":"true"}` is accepted, and tightened the validation error so valid JSON with a non-string value is described accurately. Please check. 

cc: @alyssacgoins

controllers/config/defaults.go
```diff
@@ -252,17 +252,35 @@ var allowedDBExtraParams = map[string]struct{}{
 func ValidateDBExtraParams(raw string) (map[string]string, error) {
 	var params map[string]string
 	if err := json.Unmarshal([]byte(raw), &params); err != nil {
```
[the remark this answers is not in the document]
alyssacgoins: @ugiordan this as well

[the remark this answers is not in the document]
ugiordan: Well, this is unrelated to the changes, but it looks like a bug that was already there :D Fixed it.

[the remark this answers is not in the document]
ugiordan: @coderabbitai please check. I fixed it.

alyssacgoins:
/lgtm
@mprahl can I get your thought on this?

api/v1/dspipeline_types.go
```diff
@@ -304,8 +304,11 @@ type Database struct {
 	// when opening a connection with the Database.
 	// ref: https://github.com/go-sql-driver/mysql?tab=readme-ov-file#dsn-data-source-name
 	//
-	// Value must be a JSON string. For example, to disable tls for Pipeline Server DB connection
-	// the user can provide a string: {"tls":"true"}
+	// Value must be a JSON string. For example, to enable TLS for the Pipeline Server DB connection,
+	// the user can provide a string: {"tls":"true"}. The insecure "skip-verify" and "preferred"
```
mprahl, line 308: @ugiordan , I have no idea if customers are using this but it doesn't seem right to break on upgrades for customers that opted in to a less secure connection. Where is this requirement coming from?

controllers/config/defaults.go
```diff
@@ -252,17 +252,38 @@ var allowedDBExtraParams = map[string]struct{}{
...
 	return params, nil
 }
 
+// ValidateDBTLSMode rejects MySQL TLS modes that bypass certificate verification.
+func ValidateDBTLSMode(mode string) error {
+	if strings.EqualFold(mode, "skip-verify") || strings.EqualFold(mode, "preferred") {
+		return fmt.Errorf(
+			"database TLS mode %q disables certificate verification and is rejected for security compliance; "+
+				"use tls=true and configure spec.apiServer.cABundle for self-signed or internal CA certificates; "+
+				"see https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/docs/database-tls.md",
```
mprahl, line 280: I don't think we should be pointing customers to a GitHub markdown file. Include the details here or omit the link.

controllers/config/defaults.go
```diff
@@ -252,17 +252,38 @@ var allowedDBExtraParams = map[string]struct{}{
...
+			if err := ValidateDBTLSMode(params[key]); err != nil {
+				return nil, err
+			}
+		}
 	}
 	return params, nil
 }
 
+// ValidateDBTLSMode rejects MySQL TLS modes that bypass certificate verification.
+func ValidateDBTLSMode(mode string) error {
```
mprahl, line 275: Do we also need to disallow `PodToPodTLS` to `false`?

api/v1/dspipeline_types.go
alyssacgoins: @mprahl this requirement is a part of the TLS Compliance work - https://redhat.atlassian.net/browse/RHOAIENG-94348 Jira under the https://redhat.atlassian.net/browse/RHOAIENG-90692 epic. Ugo is helping us get this in for EA2

controllers/database.go
```diff
@@ -139,6 +139,9 @@ var ConnectAndQueryDatabase = func(
 	defer cancel()
 
 	var tlsConfig *cryptoTls.Config
+	if err := config.ValidateDBTLSMode(tls); err != nil {
```
mprahl, line 142: Should the builtin MariaDB be configured with `require_secure_transport`?

controllers/config/defaults.go
ugiordan: Agreed. I removed the GitHub link and made the error self-contained. It now gives the exact replacement, `customExtraParams` set to `{"tls":"true"}`, and tells users to configure `spec.apiServer.cABundle` for self-signed or internal CA certificates.

api/v1/dspipeline_types.go
ugiordan: Yes, `tls=false` is still supported. The validation only rejects `skip-verify` and `preferred`, and the runtime path still explicitly handles `tls=false` by not configuring TLS.

The requirement here is specifically to remove the certificate-verification bypass reported by the centralized scanners, not to require TLS for every database connection.

controllers/config/defaults.go
ugiordan: It is not required for this scanner finding, but I agree it is worth reviewing as separate security hardening.

Disallowing `podToPodTLS=false` would change an existing public option and could affect Kubernetes and development deployments, so I would rather handle that as a separate product and security decision instead of adding it to this fix.

controllers/database.go
ugiordan: Same as before, enabling require_secure_transport when podToPodTLS is enabled seems like a useful defense-in-depth measure. It would prevent the built-in MariaDB from accepting an accidental plaintext connection while TLS is configured.

I would track that as a separate hardening change because it modifies the MariaDB server policy rather than fixing the client-side certificate-verification bypass covered by this PR.

controllers/config/defaults.go
mprahl: @ugiordan that clarification helps! It's still a breaking change that an admin could potentially run into. It's unlikely, but how are we notifying customers of this change across the platform?

I personally would rather not go forward with this considering there is very little to gain but if it's an absolute requirement, I can be overruled.

ugiordan: That’s fair. We could leave the current behavior unchanged and follow the approved TLS scanner exception process described here:

https://github.com/opendatahub-io/security-config/tree/main#tls-workflows

I agree that this remains a security concern, since `skip-verify` disables certificate verification and `preferred` can fall back to plaintext. I’ll leave it up to you to decide whether to fix the behavior or accept the risk through that process.

mprahl: Since TLS can be completely disabled in the current PR, I don't think much is gained.

My vote is to do the exception and in this PR, log an error or warning when TLS is not on or not set to required.

ugiordan: Agreed. However, I wouldn't add an error for `tls=false`, as that would contradict the goal of maintaining compatibility. A warning is reasonable, but the exception and warning shouldn't be presented as a solution to the security issue.

ugiordan: Updated.

mprahl (approved):
/lgtm
/approve

PR #1104, opened 2026-09-14
feat(controller): add AIPipelines modular mode

lugi0:
## The issue resolved by this Pull Request:

Resolves [RHOAIENG-91564](https://redhat.atlassian.net/browse/RHOAIENG-91564)

## Description of your changes:

Adds the opt-in controller path for the `AIPipelines` platform module:

- Registers an `AIPipelines` reconciler behind the `DSPO.EnableAIPipelinesModuleController` feature flag, which defaults to disabled.
- Watches the singleton `AIPipelines` resource and all DSPA instances.
- Aggregates DSPA and Argo Workflows controller readiness into the platform module status.
- Validates the configured Argo Workflows management state and reports configuration, component, and overall readiness conditions.
- Preserves condition transition times and reports the resolved platform release.
- Adds the required RBAC, deployment configuration, and unit coverage.

The existing DSPA controller remains active in both modes, allowing the platform ownership handoff to be enabled in a coordinated follow-up.

## Testing instructions

Run:

```shell
make unittest
```

Verify that:

- The modular controller is disabled by default and can be enabled through operator configuration.
- DSPA status changes enqueue reconciliation of the singleton `default-aipipelines` resource.
- Module readiness reflects DSPA and Argo Workflows readiness and the configured management state.
- Existing unit tests continue to pass.

## Checklist

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work.

[RHOAIENG-91564]: https://redhat.atlassian.net/browse/RHOAIENG-91564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Added an optional AI Pipelines controller that monitors pipeline resources and DSPA readiness.
  - Reports overall AI Pipelines status, readiness, phase, and platform release information.
  - Aggregates component readiness and preserves status transition timing.
  - Added permissions required to update AI Pipelines status.

- **Configuration**
  - Added a feature flag for enabling the AI Pipelines controller.
  - The controller remains disabled by default; existing ownership behavior continues unless explicitly enabled.

- **Tests**
  - Added coverage for readiness reporting, configuration behavior, error handling, and reconciliation outcomes.

PR #1105, opened 2026-09-15
feat(controller): add Argo lifecycle controller

lugi0:

Resolves [RHOAIENG-91565](https://redhat.atlassian.net/browse/RHOAIENG-91565)

Adds lifecycle management for the platform-owned Argo Workflows controller in modular mode:

- Registers a dedicated reconciler for the singleton `AIPipelines` resource.
- Applies the Argo Workflows controller assets and required RBAC when the component is managed.
- Removes managed resources when requested while preserving shared workflow CRDs.
- Avoids adopting workflow CRDs that are not owned by the platform module.
- Uses the module's Argo configuration when reconciling supported DSPA instances.
- Reports Argo and overall module readiness, and gates platform release updates until provisioning succeeds.
- Adds unit coverage for reconciliation, ownership, configuration, readiness, and failure behavior.

This is the next change after #1104 in the AIPipelines modularization stack.

## Testing instructions

Run:

```shell
make unittest
```

Verify that:

- Managed mode creates the Argo Workflows controller assets.
- Removed mode cleans up managed assets without deleting shared workflow CRDs.
- Foreign workflow CRDs are not adopted.
- Module readiness and platform release reflect provisioning success and component state.
- Existing unit tests continue to pass.

## Checklist

- [x] The developer has manually tested the changes and verified that the changes work.

[RHOAIENG-91565]: https://redhat.atlassian.net/browse/RHOAIENG-91565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Added automatic installation and management of shared Argo Workflows resources for AI Pipelines.
  - AI Pipelines configuration takes precedence over legacy Argo configuration in modular mode.
  - Configuration changes now refresh supported data science pipelines automatically.
  - AI Pipelines now reports readiness based on deployment and Argo resource health.

- **Bug Fixes**
  - Improved recovery from resource drift and incomplete or failed manifest application.
  - Prevented adoption of workflow definitions owned by another installation.
  - Improved cleanup and readiness handling for shared Argo resources.


controllers/aipipelines_argo_controller.go
```diff
@@ -0,0 +1,449 @@
...
+
+func (r *AIPipelinesArgoReconciler) checkWorkflowCRDOwnership(ctx context.Context) error {
+	crd, err := r.getWorkflowCRD(ctx)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("check existing Argo Workflow CRD: %w", err)
+	}
+	if crd.GetLabels()[legacyPipelinesComponentLabel] == "true" || crd.GetAnnotations()[argoManagedAnnotation] == "true" {
```
VaniHaripriya: `checkWorkflowCRDOwnership() `rejects `workflows.argoproj.io` unless it already has our ownership labels/annotations. Existing DSPO installs deploy that CRD from config/argo without those markers, so enabling modular Managed mode on an existing cluster would fail here. Was that intentional, or should we adopt the CRD when its spec matches our embedded asset?

controllers/aipipelines_argo_controller.go
```diff
@@ -0,0 +1,449 @@
...
+
+	if len(missing)+len(drifted)+len(remaining) > 0 {
+		items := append(append(missing, drifted...), remaining...)
+		sort.Strings(items)
+		return argoLifecycleObservation{Status: metav1.ConditionFalse, Reason: "ArgoResourcesProgressing", Message: "Waiting for shared Argo resources: " + strings.Join(items, ", ")}
+	}
+	if state == common.Removed {
+		return argoLifecycleObservation{Status: metav1.ConditionTrue, Reason: "Removed", Message: "Shared Argo controller resources are removed; CRDs are preserved"}
+	}
+	return argoLifecycleObservation{Status: metav1.ConditionTrue, Reason: "ArgoResourcesReady", Message: "Shared Argo resources are ready"}
```
alyssacgoins, line 432: [P2] Argo reports ready before CRDs are established

 `observeArgoLifecycle` reports `ArgoResourcesReady` based on CRD existence and spec alone. Kubernetes may persist a newly created CRD while its Established condition is still False, causing the module to publish successful readiness, provisioning, and release status before the API is usable. Consider requiring `Established=True` here for each managed CRD before reporting Argo ready.

controllers/aipipelines_argo_controller.go
```diff
@@ -0,0 +1,449 @@
...
+		controllerutil.AddFinalizer(updated, argoLifecycleFinalizer)
+		if err := r.Update(ctx, updated); err != nil {
+			return ctrl.Result{}, fmt.Errorf("add AIPipelines Argo finalizer: %w", err)
+		}
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	switch module.Spec.ArgoWorkflowsControllersManagementState() {
+	case common.Managed:
+		if err := r.checkWorkflowCRDOwnership(ctx); err != nil {
```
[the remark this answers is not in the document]
alyssacgoins: @lugi0 +1 here - 

This checks ownership only for workflows.argoproj.io, then reconciles every other named asset without an ownership check. If the workflow CRD is absent, an unrelated same-name resource—such as argo-cluster-role—is adopted, overwritten, and assigned the module owner reference, making it eligible for deletion with the module. Can we require the legacy label or managed annotation on each existing asset before modifying it?

alyssacgoins (changes requested):

controllers/aipipelines_argo_controller.go
lugi0: Thanks for flagging this. The ownership check is intentional and does not block the supported ODH upgrade path. The existing ODH component handler renders DSPO’s Argo resources with app.opendatahub.io/data-science-pipelines-operator=true, including the Workflow CRD. The modular reconciler accepts that legacy label and adds its new managed annotation during reconciliation.
Currently, an unlabeled CRD is intentionally treated as foreign; I am ok with changing this if my understanding here is incorrect.

controllers/aipipelines_argo_controller.go
lugi0: Addressed in [c554990](https://github.com/opendatahub-io/data-science-pipelines-operator/commit/c554990). Argo readiness now requires every managed CRD to report `Established=True`; otherwise the module remains progressing. I also added regression coverage for the `Established=False` to `True` transition.

controllers/aipipelines_argo_controller.go
```diff
@@ -0,0 +1,498 @@
...
+			}
+			if !established {
+				drifted = append(drifted, desired.GetKind()+"/"+desired.GetName())
+				continue
+			}
+		}
+		if current.GetAnnotations()[argoManagedAnnotation] != "true" ||
+			current.GetLabels()[legacyPipelinesComponentLabel] != "true" ||
+			current.GetLabels()[config.DSPVersionk8sLabel] != config.DSPV2VersionString ||
+			!argoManagedFieldsMatch(current, desired) {
```
alyssacgoins, line 419: `ownerReferences` are reconciled above but excluded from this readiness comparison. If the reference is removed or replaced—and reconciliation cannot repair it—the module can still report `ArgoResourcesReady=True`. Could we include the expected owner reference in drift detection for non-CRD assets?

controllers/aipipelines_argo_controller.go
```diff
@@ -0,0 +1,498 @@
...
+}
+
+// +kubebuilder:rbac:groups=components.platform.opendatahub.io,resources=aipipelines,verbs=get;list;watch;patch;update
+// +kubebuilder:rbac:groups=components.platform.opendatahub.io,resources=aipipelines/finalizers,verbs=update
+// +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=create;list;watch
+// +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,resourceNames=applications.app.k8s.io;clusterworkflowtemplates.argoproj.io;cronworkflows.argoproj.io;viewers.kubeflow.org;workflowartifactgctasks.argoproj.io;workfloweventbindings.argoproj.io;workflows.argoproj.io;workflowtaskresults.argoproj.io;workflowtasksets.argoproj.io;workflowtemplates.argoproj.io,verbs=get;update;patch
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings;clusterroles;clusterrolebindings,verbs=create;list;watch
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings;clusterroles;clusterrolebindings,resourceNames=argo;argo-aggregate-to-admin;argo-aggregate-to-edit;argo-aggregate-to-view;argo-binding;argo-cluster-role;ds-pipeline-argo-binding,verbs=get;update;patch;delete
```
alyssacgoins: Minor RBAC accuracy issue: the managed Role is named argo-role, not argo, and this controller never calls Patch. Could we replace argo with argo-role, remove patch from this marker, and regenerate config/rbac/role.yaml? The existing broad DSPA permissions currently mask the name mismatch.

alyssacgoins (changes requested):

config/argo/assets.go
```diff
@@ -0,0 +1,103 @@
...
+					break
+				}
+				return nil, fmt.Errorf("decode embedded Argo asset %s: %w", name, err)
+			}
+			if len(object.Object) == 0 {
+				continue
+			}
+			if object.GetNamespace() != "" {
+				object.SetNamespace(namespace)
+			}
```
Al-Pragliola: Could we also set the namespace on namespaced resources that don't specify one in the YAML? `role.argo.yaml` has no namespace-it normally get one from Kustomize, but this controller reads the embedded YAML directly. That leaves argo-role with an empty namespace, and the real client will fail during the ownership check with an empty namespace may not be set when a resource name is provided. The fake client accepts this, so the current tests don't catch it

config/rbac/role.yaml
```diff
@@ -277,6 +312,26 @@ rules:
...
+  resources:
+  - clusterrolebindings
+  - clusterroles
+  - rolebindings
+  - roles
+  verbs:
+  - delete
+  - get
+  - patch
+  - update
```
Al-Pragliola, line 359: I think we need to extend the operator's permissions to cover the roles it now creates. For example, `argo-aggregate-to-admin` grants access to workfloweventbindings, but neither manager-role nor manager-argo-role has the permission. Kubernetes won't let the operator create a role granting permissions it doesn't hold unless it has escalate, so provisioning would fail with the RBAC shipped here. Could we cover the missing permissions and test this using the operator's service account?

Al-Pragliola (changes requested):

config/rbac/role.yaml
lugi0: Addressed in 1f0261b. The generated manager RBAC now covers the permissions granted by the embedded Argo roles, including `workfloweventbindings` and the other aggregate-role resources. I also added a regression test that verifies the union of ClusterRoles bound to the `controller-manager` service account covers every permission in each embedded Role and ClusterRole.

config/argo/assets.go
lugi0: Addressed in 1f0261b. `Objects` now assigns the requested namespace to every namespaced embedded asset kind, including `argo-role` even though its YAML omits `metadata.namespace`. The asset test now verifies namespace assignment for all namespaced assets and confirms cluster-scoped assets remain unset.

controllers/aipipelines_argo_controller.go
lugi0: Addressed in b2d1729. The marker now uses `argo-role`, drops the unused `patch` verb, and `config/rbac/role.yaml` was regenerated.

controllers/aipipelines_argo_controller.go
lugi0: Addressed in b2d1729. `observeArgoLifecycle` now treats a missing or replaced controller owner reference on non-CRD assets as drift, and the regression test verifies readiness becomes false until reconciliation restores the reference.

controllers/aipipelines_argo_controller.go
VaniHaripriya: Thank you @lugi0 , agree with you.

alyssacgoins (approved):

PR #1110, opened 2026-09-17
fix(aipipelines): reconcile modular monitoring resources

lugi0:
Follow-up to opendatahub-operator#4089.

- reconcile the `monitoring.rhobs/v1` PrometheusRule from the DSPO AIPipelines controller
- exclude the PrometheusRule from platform-provided module manifests while retaining it for standalone deployments
- watch for the PrometheusRule CRD and tolerate clusters where it is unavailable
- periodically requeue while AIPipelines is NotReady so missed dependency events converge
- add controller and embedded-manifest unit coverage

## Testing instructions
Run:
- `go test -tags=test_unit ./controllers ./config/prometheus`

Verified standalone overlays include the PrometheusRule and module overlays do not.

## Checklist
- [x] Testing instructions have been added in the PR body.

## Summary by CodeRabbit

- **New Features**
  - Added automatic creation and management of Prometheus monitoring rules for Data Science Pipelines.
  - Supported deployment configurations now include monitoring and Argo resources.
  - Monitoring rules receive the correct namespace labels and ownership information.

- **Bug Fixes**
  - Existing module-owned monitoring rules are repaired when definitions differ or are deleted.
  - Monitoring integration remains optional when the monitoring API is unavailable.
  - Incomplete pipeline resources are rechecked automatically until ready.
  - Monitoring permissions are restricted to required resources.
  - Monitoring rules owned by another controller are not modified.
  - Argo resources now require their associated CRDs to be available before reconciliation.


config/prometheus/assets.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	_ "embed"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+//go:embed datasciencepipelines-prometheusrules.yaml
```
mprahl, line 29: Do we need to add `COPY config/prometheus/ config/prometheus/` to the Dockerfile?

controllers/aipipelines_controller.go
```diff
@@ -101,12 +113,25 @@ func (r *AIPipelinesReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	})
+	enqueueModule := handler.EnqueueRequestsFromMapFunc(func(_ context.Context, _ client.Object) []reconcile.Request {
+		return []reconcile.Request{{NamespacedName: types.NamespacedName{
+			Name: aipipelinesv1alpha1.AIPipelinesInstanceName,
+		}}}
+	})
+
 	b := ctrl.NewControllerManagedBy(mgr).
 		For(&aipipelinesv1alpha1.AIPipelines{}).
+		Watches(prometheusRuleCRD, enqueueModule, builder.WithPredicates(prometheusRuleCRDFilter)).
```
mprahl: Controller predicates are evaluated after objects enter the informer cache, so this still lists/watches and caches every CRD cluster-wide just to observe one CRD which is memory intensive. `watchArgoAssets` has the same issue. Can we avoid the broad CRD informer or scope the cache/watch to the specific required CRDs?

Since you need to watch more than one CRD and CRD list/watches can only filter by one field selector, you can create a watcher per CRD and add it to the controller manager. You can see an example of this in the MLflow operator:
https://github.com/opendatahub-io/mlflow-operator/blob/388e862905abcdc12fd17e4879920033f1a52a27/cmd/main.go#L463-L479

This also allows us to scope down the RBAC required by this operator to just `list` and `watch` those CRDs if desired.

controllers/aipipelines_monitoring.go
```diff
@@ -0,0 +1,95 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const prometheusRuleCRDName = "prometheusrules.monitoring.rhobs"
+
+// +kubebuilder:rbac:groups=monitoring.rhobs,resources=prometheusrules,verbs=get;list;watch;create;update;patch;delete
```
mprahl: Could you set the `resourceName` to scope the RBAC to just the singleton it manages?

mprahl (changes requested):
Left a few comments but looks good.

mprahl (approved):

PR #1113, opened 2026-09-17
fix(mlflow): ignore missing CRD during endpoint lookup

mprahl:
Treat the missing MLflow CRD as an unavailable optional feature instead of
a reconcile error. This skips MLflow API server plugin configuration on
clusters where mlflow.opendatahub.io/v1 is not installed.

A regression test covers the missing-CRD discovery error.

Tested with: make unittest

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling when the MLflow integration is unavailable.
  - Pipeline configuration now skips MLflow settings when no endpoint is available.
  - Prevented unavailable MLflow endpoints from being cached as valid results.


alyssacgoins (commented):
/lgtm

